PR #70, opened 2026-04-02
permissions: preserve full-access policy in voice + robust Kotlin bindings on Linux

ericbosch:
## Summary
- preserve selected permission/sandbox policy through voice handoff/resume flows (iOS + Android)
- default Litter launch sandbox to danger-full-access on both platforms
- fix Linux Kotlin binding generation by sourcing UniFFI metadata from Android .so (avoid host V8 linker failure)

## Validation
- make bindings-kotlin
- make android-emulator-fast (BUILD SUCCESSFUL)
- generated APK sha256: 84ef3ff4b185b68e329bccfb5722a1be315fdaf49bee7ad215b17ffdf55b0b78

## Notes
- no submodule commit included;  remains local-only as requested

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/state/AppLaunchState.kt
+++ b/apps/android/app/src/main/java/com/litter/android/state/AppLaunchState.kt
@@ -36,8 +36,8 @@ data class AppLaunchStateSnapshot(
 private const val PREFS_NAME = "litter.launchState"
 private const val APPROVAL_POLICY_KEY = "litter.approvalPolicy"
 private const val SANDBOX_MODE_KEY = "litter.sandboxMode"
-private const val DEFAULT_APPROVAL_POLICY = "inherit"
-private const val DEFAULT_SANDBOX_MODE = "inherit"
+private const val DEFAULT_APPROVAL_POLICY = "never"
+private const val DEFAULT_SANDBOX_MODE = "danger-full-access"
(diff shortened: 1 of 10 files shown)


dnakov (commented):
Two regressions still need attention here:

1. The voice/handoff path still does not preserve per-thread permission overrides. On iOS, `VoiceRuntimeController` now derives a launch policy from global `UserDefaults` and reuses that for resume/start-turn (`apps/ios/Sources/Litter/Models/VoiceRuntimeController.swift`, especially `startRealtimeVoiceSession`, `executeHandoffSendTurn`, and `currentLaunchPolicy`). On Android, `HandoffAction.SendTurn` uses `approvalPolicyValue()` / `turnSandboxPolicy()` without the target thread key even though `AppLaunchState` already exposes thread-aware variants. That means a thread explicitly set to read-only / on-request gets flipped back to the global default during voice handoff or resume.

2. Several iOS resume entrypoints now hard-code `.never` + `.dangerFullAccess`, which silently escalates permissions on existing threads. `AppLifecycleController.refreshTrackedThreads(...)` resumes backgrounded threads with full access, and `SubagentDetailSheet.loadThreadIfNeeded()` does the same just to open the subagent transcript. Before this patch those resume calls passed `nil`, so the existing effective policy stayed intact. After this patch, reopening a previously constrained thread changes its effective launch policy.

PR #74, opened 2026-04-07
mobile: initialize rustls provider before bridge startup

saintmob:
## Summary
- Install the process-wide rustls crypto provider from the shared mobile init path so iOS/Android bridge construction cannot hit the Rustls 0.23 provider panic.
- Keep `MobileClient::new()` as a second line of defense.
- Add the shared workspace dependency needed by the mobile Rust crate.

## Verification
- Rebuilt the iPhone target after the Rust init fix.
- Confirmed the app launches successfully on a real iPhone.

## Scope
- This PR intentionally includes only the mobile Rustls startup fix.
- It does not include unrelated local edits or the separate websocket token/logging follow-up.


saintmob:
IMG_7822
IMG_7917

PR #75, opened 2026-04-13
expanded resolver ssh probe

ryanchen01:
## Summary

This updates the POSIX SSH resolver in `codex-mobile-client` so remote Codex binary lookup covers more real-world
install locations while keeping the probe side-effect free.

## What changed

- expanded remote resolver diagnostics to report:
  - `PNPM_HOME`
  - `NVM_BIN`
  - npm prefix
  - npm/pnpm global bin directories
- added npm/pnpm-aware candidate paths for both `codex` and `codex-app-server`
- included the local package install path at `~/.litter/codex/node_modules/.bin/codex`
- switched Cargo-based lookups to respect `CARGO_HOME`
- refactored the POSIX resolver script to use shared candidate helpers instead of a long `if/elif` chain
- added regression tests for:
  - profile init behavior
  - package-manager-aware resolver generation

## Why

Remote bootstrap was too narrow about where `codex` might be installed. In practice it can come from npm/pnpm globals,
NVM-managed installs, local `.litter` installs, or non-default Cargo locations.

The follow-up cleanup in this change also avoids two risks in the new probe path:

- no `corepack pnpm ...` fallback, so resolution does not activate/download tooling as a side effect
- no extra `head -n 1` subprocesses on commands that already return a single line

## Impact

- improves SSH bootstrap reliability when Codex is installed outside the old hard-coded paths
- gives better resolver diagnostics when remote setup fails
- keeps the lookup path read-only and lower overhead

### DIFF
--- a/shared/rust-bridge/codex-mobile-client/src/ssh.rs
+++ b/shared/rust-bridge/codex-mobile-client/src/ssh.rs
@@ -1154,8 +1154,14 @@ printf '%s/codex-ipc/ipc-%s.sock' "$tmp" "$uid""#;
     async fn fetch_codex_resolver_diagnostics(&self) -> String {
         let script = format!(
             r#"{profile_init}
+{pkg_probe}
 printf 'shell=%s\n' "${{SHELL:-}}"
 printf 'path=%s\n' "${{PATH:-}}"
+printf 'pnpm_home=%s\n' "${{PNPM_HOME:-}}"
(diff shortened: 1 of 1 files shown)


dnakov:
I dont think codex-app-server is a thing

ryanchen01:
I've also removed the `$HOME/.cargo/bin/codex-app-server` originally in the codebase.

PR #80, opened 2026-04-17
ssh: detect bun-installed codex

jpenilla:
On my machine, the Bun global bin is added to `PATH` in `.zshrc`. My `.zshrc` has zsh specific/non-POSIX things in it, so it could not be used for the app to discover codex. A broader change could be to prioritize the user's configured login shell with fallback to common shells where each respective shell reads it's own configs. But it seems easier to patch the existing system with Bun support similar to how npm/pnpm is handled for now. Side note: I haven't looked closely at the resolver logic from before #75, but until recently my Bun-installed codex was detected (or at least I wasn't prompted to install it), so this may be a regression fix.


dnakov:
yeah i couldnt come up with a great system, seemed like there are fewer ways to find codex than to deal with all the shell stuff. will prob refactor this when we add support for pi and opencode

PR #83, opened 2026-04-21
fix(ios): submit composer on Return with external keyboard

andrerfneves:
## Problem
On iPad with an external keyboard, pressing **Return** in the inline message composer inserts a newline () instead of submitting the message. Users have to reach for the on-screen send button, which breaks the flow.

## Solution
Intercept hardware key events in \"PasteAwareComposerUITextView\" via \"pressesEnded(_:with:)\".

- **Return** → submits the message
- **Cmd+Return** → also submits (common muscle memory)
- **Shift+Return** → inserts a newline (preserves multi-line input ability)

## Scope
- Only the **inline entry-row composer** opts in (\"ConversationComposerEntryRowView\").
- The expanded composer and all other usages remain untouched.
- Fully backward-compatible: new parameters default to nil/false.

## Notes
- Uses iOS 13.4+ \"UIPress.key\" / \"modifierFlags\" APIs; app targets iOS 18.0.
- Zero Rust/shared-layer changes — pure iOS SwiftUI/UIKit.

### DIFF
--- a/apps/ios/Sources/Litter/Views/ConversationComposerEntryRowView.swift
+++ b/apps/ios/Sources/Litter/Views/ConversationComposerEntryRowView.swift
@@ -78,7 +78,9 @@ struct ConversationComposerEntryRowView: View {
                     ConversationComposerTextView(
                         text: $inputText,
                         isFocused: $isComposerFocused,
-                        onPasteImage: onPasteImage
+                        onPasteImage: onPasteImage,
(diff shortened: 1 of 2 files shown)


dnakov:
I just did a little iPad overhaul, should be solved in there, can u check latest testflight pls?

andrerfneves:
Yup, closing this. Thank you!

PR #94, opened 2026-04-30
fix(ios): load threads on home mount

ratley:
## Summary
Load threads when the home screen mounts instead of only after opening search

## Verification
Tested in iOS simulator

## Notes
Apparently it's the same case on Android so maybe that's intentional? If so, you can just close this


dnakov:
resolved with other commits

PR #109, opened 2026-05-10
edit saved server connections from settings

kaynansc:
## Summary
- Adds an edit flow for saved server connections in iOS Settings: tapping a row opens a sheet to rename and adjust the connection (local, direct Codex, websocket, or SSH).
- Wires save vs. save-and-reconnect paths, including SSH re-auth via the existing `SSHLoginSheet`, and surfaces failures through an alert.
- Refactors the existing add-server sheet plumbing into a single `activeServerSheet` enum to host add/edit/ssh-reconnect cases cleanly.

## Test plan
- [ ] Open Settings → tap a saved server → rename and Save; row reflects the new name without disconnecting.
- [ ] Open Settings → tap a saved server → change connection mode and Save & Reconnect; verify reconnection succeeds for local, direct Codex, and websocket modes.
- [ ] Switch a server to SSH and Save & Reconnect; confirm `SSHLoginSheet` appears and reconnect completes after credentials are provided.
- [ ] Trigger a failure path (e.g. invalid websocket URL) and confirm the "Server Update Failed" alert appears.
- [ ] Remove a saved server from Settings; confirm SSH session is closed and server is disconnected.


apps/android/app/src/main/java/com/litter/android/ui/settings/SettingsSheet.kt
```diff
@@ -456,6 +545,424 @@ private fun ServerSettingsRow(
...
+    }
+
+    fun persist(saved: SavedServer) {
+        val existing = SavedServerStore.load(context).toMutableList()
+        existing.removeAll { it.id == saved.id }
+        existing.add(saved)
+        SavedServerStore.save(context, existing)
+        appModel.reconnectController.setMultiClankerAndQuicEnabled(true)
+        appModel.reconnectController.syncSavedServers(
+            existing.filter { it.rememberedByUser }.map { it.toRecord(context) }
```
dnakov, line 726: This does not compile as submitted. `toRecord` is a top-level extension in `com.litter.android.state`, so this file needs `import com.litter.android.state.toRecord`. I verified `./gradlew :app:compileDebugKotlin` fails with `Unresolved reference 'toRecord'` here and on the later `map { it.toRecord(context) }` call; adding that import makes the Kotlin compile pass.

apps/ios/Sources/Litter/Views/SettingsView.swift
```diff
@@ -288,6 +327,568 @@ struct SettingsView: View {
...
+        )
+
+        guard reconnect else { return }
+        reconnectServer(using: configuration)
+    }
+
+    private func reconnectServer(using configuration: SettingsServerConnectionConfiguration) {
+        let server = configuration.discoveredServer
+        Task {
+            await SshSessionStore.shared.close(serverId: server.id, ssh: appModel.ssh)
```
dnakov, line 428: For SSH reconnects this disconnects the existing session before the user confirms replacement credentials. The `.ssh` branch below then opens `SSHLoginSheet`; if the user cancels that sheet, the prior connection has already been closed. Android’s implementation delays disconnect until `SSHLoginDialog.onConnect`, which preserves the current connection on cancel. iOS should either route through the same shared reconnect flow or special-case SSH so disconnect happens only after credentials are submitted.

apps/android/app/src/main/java/com/litter/android/ui/settings/SettingsSheet.kt
```diff
@@ -456,6 +545,424 @@ private fun ServerSettingsRow(
...
+                    rememberedByUser = true,
+                )
+            }
+            ServerConnectionMode.SSH -> {
+                val resolvedHost = host.trim()
+                if (resolvedHost.isEmpty()) {
+                    validationError = "Host cannot be empty."
+                    return null
+                }
+                val resolvedSSHPort = sshPort.trim().toIntOrNull()
```
dnakov, line 631: This validation is looser than the existing manual SSH entry path. Ports above `65535` pass here and later get converted with `toUShort()`, which wraps instead of rejecting. The existing `buildManualSshEntry` checks `sshPort !in 1..65535`; the editor should use the same bounds.

apps/android/app/src/main/java/com/litter/android/ui/settings/SettingsSheet.kt
```diff
@@ -456,6 +545,424 @@ private fun ServerSettingsRow(
...
+                if (resolvedSSHPort == null || resolvedSSHPort <= 0) {
+                    validationError = "SSH port must be a valid number."
+                    return null
+                }
+                val resolvedWakeMAC = SavedServer.normalizeWakeMac(wakeMAC.trim())
```
dnakov: Invalid wake MAC input is silently converted to `null`, so a typo clears the saved wake MAC instead of surfacing the validation error. The existing manual SSH flow rejects `wakeInput.isNotEmpty() && normalizedWakeMac == null`; this editor should match that behavior for parity and to avoid lossy edits.

dnakov (changes requested):
Thanks for wiring both platforms. I think this needs another pass before merge.

The main blocker is Android currently does not compile: `SettingsSheet.kt` calls the `SavedServer.toRecord` extension without importing it. I verified in an isolated PR worktree that `./gradlew :app:compileDebugKotlin` fails with `Unresolved reference 'toRecord'`; adding `import com.litter.android.state.toRecord` lets the Kotlin compile complete.

There are also two behavioral/architecture issues worth fixing now: iOS disconnects the existing SSH session before showing the credential sheet, so canceling re-auth leaves the prior connection dead, and the new Android editor validation is looser than the existing manual-entry path. The cross-platform shape would be cleaner if both platforms persist the edited record, sync `SavedServerRecord`s, then use the shared Rust reconnect path where possible, with native code limited to UI and credential prompts.

kaynansc:
Thank you for reviewing @dnakov Just addressed the comments!

dnakov:
Thanks for the update. One blocker still remains on Android: Save & Reconnect closes the edit sheet even when reconnect fails. In the failure path the catch sets validationError, but onSave() still runs afterward, clearing editTarget and removing the UI that would show the error. Please keep the sheet open on reconnect failure and only call onSave() after a successful reconnect/save. The earlier compile/import and validation items look addressed from this pass.

PR #116, opened 2026-05-11
fix: scope composer rate limits to the active thread's CLI runtime

kaynansc:
## Summary
- Composer rate-limit badges were always reading from a single per-server snapshot, so on a multiplexed (Alleycat) session opening a Claude Code thread still showed Codex usage at the bottom of the composer.
- Track rate limits keyed by `(server_id, runtime_kind)` so the composer can look up the snapshot for the thread's own runtime — Claude Code threads show Claude usage, Codex threads show Codex usage, and runtimes that have not reported usage render no badge.
- Plumb `runtime_kind` through `EventProcessor::process_notification` and `UiEvent::AccountRateLimitsUpdated` so the reducer knows which runtime owns each update; expose `rate_limits_by_runtime` on the UniFFI boundary as a `Vec<AppRateLimitsForRuntime>` (UniFFI does not support enum-keyed maps).

## Test plan
- [x] `make rust-test` (662 passed)
- [x] `make ios-sim-fast` (BUILD SUCCEEDED)
- [ ] Android emulator build — skipped locally (no Android NDK/JDK17 installed on this machine). Kotlin change is a one-liner analogous to the iOS wiring and is covered by the shared Rust tests for the storage/event-routing change.
- [ ] Manual: on a server with both Codex and Claude Code runtimes (Alleycat), open a Claude Code thread and confirm the bottom-of-composer rate-limit badges either show Claude Code's usage or render no badge (depending on whether the runtime has reported usage), and never show Codex's usage.


dnakov:
This is still blocked on conflicts with current main, especially around shared Rust session event handling. Please rebase onto main and keep the rate-limit state scoped through the Rust-owned active thread/runtime boundary. After rebasing, rerun at least the shared Rust checks for codex-mobile-client; I can re-review once the conflict resolution is visible.

PR #117, opened 2026-05-11
[codex] Fix thread-scoped input prompts

kaynansc:
## Summary
- keep dismissed pending-user-input prompts hidden across iOS conversation view remounts
- scope Android pending-user-input overlays to the currently open thread route instead of showing every pending request globally
- update the iOS preview fixture for the current `AppServerSnapshot` shape

## Root cause
Pending user-input prompts were tracked too locally on iOS and too globally on Android. That could make dismissed prompts reappear after composer remounts and could show pending requests outside the thread that originated them.

## Validation
- `xcodebuild -project apps/ios/Litter.xcodeproj -scheme Litter -configuration Debug -destination 'platform=iOS Simulator,name=iPhone 17 Pro' build`\n- `JAVA_HOME=/opt/homebrew/opt/openjdk@17 ANDROID_HOME=$HOME/Library/Android/sdk ANDROID_SDK_ROOT=$HOME/Library/Android/sdk make android-debug`\n\n## Notes\n- This PR now contains only `0f113c9` (`fix thread-scoped input prompts`).\n- `make ios-sim-fast` still stops in the Rust prep step because `patches/codex/realtime-webrtc-env-apikey.patch` does not apply cleanly to the current dirty Codex submodule; the direct Xcode build was used after confirming existing Rust simulator artifacts were present.\n


dnakov:
This needs a rebase before it can be reviewed/merged. When resolving the Android side, please preserve current main’s thread relevance behavior rather than narrowing prompts to exact thread-id equality only; current UI paths intentionally tolerate empty/missing thread ids for compatibility. Also keep the MCP elicitation approval filtering that landed on main. After that, rerun the Android unit/compile checks.

PR #148, opened 2026-05-19
Fix Windows npm installer extraction

yashau:
## Summary
- Patch cargo-dist's generated npm package before publishing so the Windows installer prefers `tar -xf` for zip extraction.
- Add a stricter PowerShell fallback with `-ExecutionPolicy Bypass` and `$ErrorActionPreference = "Stop"`.
- Verify expected binaries exist before the installer reports success.

## Why
On Windows, the current npm package can print `kittylitter has been installed!` even when extraction failed and `.bin_real/kittylitter.exe` is missing. I hit this with `npx kittylitter` on Windows: `Expand-Archive` failed to autoload `Microsoft.PowerShell.Archive`, but `powershell.exe` exited 0, so the wrapper then failed with `spawnSync ... .bin_real\kittylitter.exe ENOENT`.

## Verification
- `node --check tools/scripts/patch-kittylitter-npm-package.js`
- `npx -y prettier@3.8.1 --check tools/scripts/patch-kittylitter-npm-package.js .github/workflows/release.yml`
- Patched the published `kittylitter-0.3.2.tgz` locally and confirmed the generated `binary-install.js` contains the Windows `tar` path and missing-binary guard.
- Installed the patched tarball in a throwaway Windows npm project and confirmed `node_modules/kittylitter/node_modules/.bin_real/kittylitter.exe` exists and runs.
- Confirmed the stricter PowerShell fallback extracts the release zip successfully on the same Windows machine.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -314,6 +314,7 @@ jobs:
       PLAN: ${{ needs.plan.outputs.val }}
     if: ${{ !fromJson(needs.plan.outputs.val).announcement_is_prerelease || fromJson(needs.plan.outputs.val).publish_prereleases }}
     steps:
+      - uses: actions/checkout@v5
       - name: Fetch npm packages
         uses: actions/download-artifact@v7
         with:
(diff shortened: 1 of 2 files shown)


0xSero (changes requested):
The Windows fix is still needed, and I confirmed the patcher applies successfully to the official cargo-dist 0.31.0 npm package. This cannot merge as currently wired, though.

First, rebase while preserving current main’s Node 24 setup and `package-manager-cache: false`.

Second, the patch currently runs in `publish-npm` after `host` has already uploaded/released the original tarball. That makes the npm-published artifact differ from the GitHub release artifact and dist provenance/checksums. Please apply the transformation before the artifact enters both publication paths, or regenerate and verify the manifest/checksums from the final bytes.

Please also add a pinned cargo-dist fixture test and a Windows CI smoke test that installs the final repacked tarball, executes `kittylitter.exe`, and verifies the missing-binary path fails rather than printing success.

yashau:
Unfortunately I do not use this software anymore. Hopefully someone else can do this for you.

0xSero:
I preserved the Windows extraction fix and replaced this stale/conflicted branch with current-main PR #210. The replacement keeps Node 24, finalizes the npm tarball before either publication path, rewrites and verifies the dist manifest plus unified checksum, publishes npm provenance, and adds a bounded Windows executable smoke test. Local fixture and final-artifact validations pass. Closing this PR as superseded by #210; thank you for isolating the original Windows failure.

PR #157, opened 2026-05-25
fix: bridge iOS skills and add unsigned IPA build

NightVibes33:


NightVibes33:
Sorry I was trying to push to my main branch not urs 🙄 😆

PR #158, opened 2026-05-25
Preserve realtime item update boundaries

lum3on:
## Summary
- preserve important `ThreadItemChanged` boundaries for operation-like conversation items instead of coalescing them into only the latest item
- keep existing coalescing behavior for ordinary non-tool item updates
- add focused tests for tool lifecycle, tool progress, file-change, and multi-agent progress boundaries

## Why
Fast realtime updates can otherwise collapse before the iOS client sees intermediate tool/progress states. This keeps status/progress-bearing item transitions visible for any backend that emits them through the app-store subscription path.

## Validation
- `git diff --check`
- Not run: `cargo fmt --check -p codex-mobile-client` / Cargo tests, because this local Mac shell does not have `cargo` installed.

### DIFF
--- a/shared/rust-bridge/codex-mobile-client/src/ffi/app_store.rs
+++ b/shared/rust-bridge/codex-mobile-client/src/ffi/app_store.rs
@@ -1,6 +1,7 @@
 //! UniFFI-exported state surface for the canonical Rust app store.
 
 use crate::MobileClient;
+use crate::conversation_uniffi::{HydratedConversationItem, HydratedConversationItemContent};
 use crate::ffi::ClientError;
 use crate::ffi::shared::{blocking_async, shared_mobile_client, shared_runtime};
 use crate::store::{AppSnapshotRecord, AppStoreUpdateRecord, AppThreadSnapshot};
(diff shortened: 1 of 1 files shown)


lum3on:
A bit more context on this PR:

This is the generic iOS/client-side part of the realtime progress fix. The issue is that fast progress updates can arrive
as multiple `ThreadItemChanged` events for the same conversation item, but the app-store subscription currently coalesces those into the latest item before iOS sees them.

That is fine for normal text-like updates, but it loses important intermediate states for tools and progress-bearing items,
such as tool started/running/completed, MCP progress messages, file changes, multi-agent/subagent state changes, web
search, and image generation.

This PR keeps normal coalescing in place, but preserves boundaries when the item represents meaningful operation progress. That makes the fix general: any Pi harness or backend that emits progress through the normal Litter update path can benefit from it.

lum3on:
@dnakov if possible pls give a heads up once feature is shipped. If you want me to test with one of my harnesses pls let me know. You could then add me to test dev so I can plug and confrim that it works e2e

dnakov:
its shipped automatically already to testflight and android

PR #178, opened 2026-07-05
fix(android): reject stale shell backlog + always-visible keyboard toggle

wahaj4311:
## Summary
1. **Terminal breakage on reconnect** (\`v1-Backend: terminal JSON-RPC stream closed\`): the shared \`remote_alleycat\` read loop was accepting replayed backlog frames from a prior killed shell session while its own \`shell_session_id\` was still \`None\`. A stale \`shell/exit\` in the ring then broke the loop before \`initialize\` returned. Guard by treating an unset expected id as reject-until-spawned instead of accept-all.
2. **Keyboard never appearing on Android**: \`onSingleTapConfirmed\` returned false when \`terminalRenderer\` was null, so tapping the black surface during startup never called \`showIme()\`. Users had no other way to summon the IME. Fix by popping the IME on every confirmed tap, and add an \`Icons.Outlined.Keyboard\` \`IconButton\` in the terminal header wired to a new file-local \`showTerminalKeyboard()\` helper (via a \`WeakReference\` to the active \`GhosttyAndroidSurfaceView\`).

Pairs with dnakov/alleycat#PR (server-side session-replay fix for the shell agent). Either patch alone fixes the reconnect breakage; both together are the belt-and-suspenders.

## Test plan
- [x] Reject-on-null-session verified: Android app opens terminal, backgrounds, reconnects — every attach is Fresh with the server-side fix and no stale exit interrupts \`initialize\`.
- [ ] Manual: tap black viewport during startup — keyboard appears.
- [ ] Manual: tap the new keyboard icon in the terminal header — keyboard appears.
- [ ] Manual: reconnect after wifi flip — keyboard button still summons IME.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/ui/terminal/GhosttySurfaceView.kt
+++ b/apps/android/app/src/main/java/com/litter/android/ui/terminal/GhosttySurfaceView.kt
@@ -431,7 +431,12 @@ private class GhosttyAndroidSurfaceView(
             }
 
             override fun onSingleTapConfirmed(e: MotionEvent): Boolean {
-                val renderer = terminalRenderer ?: return false
+                // ponytail: pop the IME on every confirmed tap, even before
+                // the renderer binds. The prior guard returned false when
+                // terminalRenderer was null, so tapping the black surface
+                // during startup never summoned the keyboard.
+                showIme()
+                val renderer = terminalRenderer ?: return true
                 if (currentSelectionRange() != null) {
                     clearSelection()
                     return true
@@ -444,7 +449,6 @@ private class GhosttyAndroidSurfaceView(
                     runCatching { context.startActivity(intent) }
                     return true
                 }
-                showIme()
                 return true
             }
         },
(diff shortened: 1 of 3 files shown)


0xSero (changes requested):
Two correctness blockers remain. First, changing `None => false` rejects stale replay but can also drop fresh `shell/output` arriving after the `shell/spawn` response is read and before `open()` records the returned session ID. Please buffer pre-spawn notifications or set/filter the ID atomically in the reader path, with tests for stale-exit rejection and immediate fresh-output preservation. Second, the process-global `WeakReference` keyboard handle can target another retained terminal surface; use a per-screen callback/handle or identity-guarded lifecycle clearing, with coverage. Please also fix the companion link to 0xSero/alleycat#28; that PR is still open and substantially behind the pinned Alleycat revision. The transport and keyboard changes should be split for safe review.

0xSero:
Closing after the full cleanup review because this combined patch is not safe to merge or useful to keep as an active queue item.

The replay change can drop fresh shell output in the spawn/registration window; the process-global weak keyboard handle can target a different retained terminal surface; the companion 0xSero/alleycat#28 branch is substantially behind; and both keyboard acceptance checks remain unverified. The earlier changes-requested review records the exact required fixes.

If either defect is reproduced on current main, please reopen it as a focused issue/PR with (1) stale-exit rejection plus immediate-fresh-output coverage, or (2) a screen-owned keyboard handle plus physical Android verification. The head branch remains available as reference.

PR #179, opened 2026-07-06
Fix Photos image preview paths

benthecarman:
## Summary
- Treat Photos library internal paths as transient iOS-only paths, not durable file attachments.
- Prevent restored image-view cards from auto-resolving `.photoslibrary` derivative paths on app open.
- Add regression coverage for Photos-library path detection and image fallback behavior.

## Verification
- `make ios-device-fast` reached Xcode but failed due to missing signing profiles/accounts for the app, live activity, watch app, and complications targets.
- Focused simulator test was attempted after installing watchOS 26.2, but simulator Rust artifact generation failed because the disk filled up (`No space left on device`).

### DIFF
--- a/apps/ios/Sources/Litter/Models/ConversationAttachmentSupport.swift
+++ b/apps/ios/Sources/Litter/Models/ConversationAttachmentSupport.swift
@@ -24,7 +24,7 @@ enum ConversationAttachmentSupport {
         .png,
         .jpeg,
         .gif,
-    ] + [UTType(filenameExtension: "webp")].compactMap { $0 }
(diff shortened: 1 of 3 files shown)


0xSero (approved):
Reviewed in an isolated current-main synthetic merge. The change remains scoped to iOS attachment/path handling, and all eight focused `ConversationAttachmentSupportTests` pass on the iPhone 17 Pro simulator. Coverage includes unreadable image fallback, Photos-library internal paths, ordinary file attachments, and image input preparation. Approved for squash merge.

PR #184, opened 2026-07-20
android: retry model picker metadata

aj47:
## What changed

- Retry model metadata loading when the server publishes an empty model list.
- Retry model and rate-limit loading when the home model picker is opened, including after connection-state changes.

## Why

An empty model list can come from a transient runtime or `model/list` failure. Android treated that empty result as cached metadata, leaving the picker stuck on `Loading models...` until a later lifecycle event.

## Validation

- `./gradlew :app:compileDebugKotlin --no-daemon`
- `./gradlew :app:testDebugUnitTest --no-daemon`
- `cargo test -p codex-mobile-client ffi::client::tests --lib`
- Verified model loading and selection on a connected Android device.


0xSero:
Thanks for working on the model-picker failure. Current main now exposes explicit model-catalog errors, a force-retry path in the Android picker, and Ultra support through merged #193. This branch conflicts in both files it changes and its original symptom is resolved, so I am closing it as superseded rather than asking you to rebase obsolete UI logic.

Issue #31, opened 2026-03-16
Android remote Codex transport should support secure `wss://` reverse-proxied deployments


lsunay:
Hi,

  I tested the Android client in a real remote deployment where Codex runs privately over `ws://` and nginx exposes it publicly over `wss://`.

  The main issue is that secure reverse-proxied deployments are a very common   real-world setup, but the Android remote flow appears to assume simpler `ws://host:port` input and does not treat `wss://domain[/path][?query]` as a first-class deployment model.

  What would help upstream:

  - support full manual websocket URLs for Codex
  - support `wss://` cleanly on Android
  - preserve scheme, path, and query parameters
  - make reverse-proxy deployments behind nginx / Cloudflare easier to use

  I prepared this in my fork here for reference:

  https://github.com/lsunay/litter-nginx-compitable/tree/wss-nginx-support

  This is not meant as “the final upstream solution”, but as a practical proof that Android remote Codex connectivity works much better when `wss://` reverse-proxied deployments are handled explicitly.

  Thanks.

PR #186, opened 2026-07-22
mobile: integrate Local Studio controllers

0xSero:
## Purpose

Fix the production Local Studio integration so Litter exposes every available paired controller agent (including Codex and Local Studio), pin the controller bridge to the validated session-isolation fixes, and remove the release-workflow failure that prevented signed production builds.

## Key changes

- Stop filtering the iOS and Android model/agent picker down to only the selected Local Studio agent.
- Return all paired controller agents from the shared Rust catalog while preserving the controller's active/default selection.
- Make controller-agent attachment atomic so an attach cannot race the selected agent mapping.
- Pin AlleyCat to merged production commit `417f2a9fe38cbed63754f0af7df61f32ec3034e6`.
- Use one temporary stable probe identity in the production E2E harness; delete it on exit and keep it out of evidence artifacts.
- Preserve controller-scoped Local Studio sessions, Codex sessions, ordered reasoning/content/tool events, files, compaction, reconnect, and authoritative hydration.
- Make R2/sccache acceleration genuinely optional in shared, Android, iOS, and Mac release jobs. Missing cache credentials now fall back to normal Rust builds instead of constructing an invalid S3 endpoint and aborting every release.
- Include current `main`, including the merged background-helper fix that prevents Local Studio's Pi worker from appearing as a separate Dock app.

## Size

- 292 additions / 78 deletions
- 10 committed files

## Verification

- Exact pinned AlleyCat merge tree matches the validated feature head.
- AlleyCat focused session-isolation tests passed; full bridge suite passed (202 passed, 1 ignored). GitHub CI passed before its merge.
- Live Local Studio production protocol acceptance: **20 passed / 0 failed**.
  - controller catalog and active model
  - new and pre-existing session hydration
  - leave/return cursor and detached active-turn visibility
  - streaming lifecycle and tool-before-final ordering
  - filesystem exec/search and file creation
  - compaction lifecycle, marker hydration, and post-compaction streaming
- `cargo build --locked` for kittylitter: passed.
- `cargo check --locked -p codex-mobile-client`: passed.
- Exact combined-head Local Studio/model-picker suite: **12 passed / 0 failed**.
- `make ios-sim-fast IOS_SIM_DEVICE='iPhone 17 Pro'`: passed on exact head `2a65405d82cf6d966a31bc42151bbdcda5a62154`.
- Exact-head simulator UI acceptance: **5 passed / 0 failed** for keyboard input, display settings reachability, expanded reasoning/command/tool details, collapsed live/recent tool output, and hidden-detail mode. Result bundle: `/tmp/litter-final-head-ui.xcresult`.
- Exact-head app installed and launched on the booted iPhone 17 Pro simulator; simulator binary SHA-256 `f84a95e8900e68a01f70b1ad2aa8f2702751db1583fd3d070512b3c6b50cbc86`.
- Three cold-launch UI passes showed both Codex and Local Studio in the picker after the picker fix.
- Exact-head `make ios-device-fast`: passed. The iPhone, Live Activity, and Watch binaries were signed and embedded successfully with the configured Apple Development team/profile; deep code-sign verification passed and the device binary SHA-256 is `2cea5902944d10ff29127d37a254263db6f70bc84a92165008e58bda6347c5f1`.
- Production release failure reproduced from GitHub Actions: empty optional R2 secrets forced `sccache` to `https://s3.auto.amazonaws.com` and killed shared prep. The workflow now uses the same credential-gated fallback pattern as Mobile CI.
- Full PR diff scanned with gitleaks: no leaks.
- Workflow YAML parsing and `git diff --check origin/main...HEAD`: clean.

## Acceptance boundary

The exact combined head is installed and running in the iOS simulator and was installed and launched successfully on the paired physical iPhone 15 Pro Max. Signed TestFlight and Google Play builds are post-merge release gates; upload, store processing, review, and public availability will be verified separately rather than inferred from source, CI, simulator, or development-device success.

Generated Rust artifacts and dirty third-party submodule worktrees are excluded. Pairing JSON remains private controller credential material.


shared/rust-bridge/codex-mobile-client/src/local_studio.rs
```diff
@@ -0,0 +1,2927 @@
...
+    let result = async {
+        let manifest = match local_studio_capabilities(&mut session).await? {
+            CapabilitiesCallResult::Capabilities(manifest) => manifest,
+            CapabilitiesCallResult::Error(error) => {
+                return Ok(LocalStudioControllerLoadResult::Error { error });
+            }
+        };
+        if !manifest
+            .capabilities
+            .contains(&LocalStudioCapability::StatsRead)
```
0xSero: [P1] Fresh Profile QR pairing does not grant this capability. I traced the producer in `vllm-studio/frontend/desktop/logic/kittylitter-pairing.ts`: it only runs `kittylitter pair` and serializes the generic `{v,node_id,token,host_name,relay}` payload. In the pinned Alleycat branch, `local_studio_agent_info` reads `GrantStore::effective`, whose default is an empty capability set; the only grant path is the separate host command `kittylitter local-studio grant`. A new phone can therefore pair successfully and then fail here with `manifest does not grant stats.read`, so QR/JSON is not a self-contained Local Studio connection. Please make Profile enrollment perform an explicit, user-consented Local Studio grant tied to the paired endpoint, and add a cold-state E2E test starting with an empty grant store. Reusing a pre-granted development endpoint does not cover this.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
```diff
@@ -345,6 +404,42 @@ impl AppClient {
...
+        limit: u64,
+    ) -> Result<crate::local_studio::LocalStudioSessionListResult, ClientError> {
+        blocking_async!(self.rt, self.inner, |c| {
+            c.list_local_studio_sessions(&server_id, limit)
+                .await
+                .map_err(|error| ClientError::Transport(error.to_string()))
+        })
+    }
+
+    pub async fn continue_local_studio_session_list(
```
0xSero: [P1] The PR claims Local Studio session read/continue and agent-turn support, but the public mobile surface stops at list pagination. `local_studio_session_read` and `local_studio_agent_turn` are private functions in `local_studio.rs` with no callers, no AppClient methods, no store reconciliation, and no UI path; dashboard session rows are non-interactive text. The app therefore cannot attach to/read a Local Studio Pi session or send the claimed agent turn. Either expose and consume typed read/turn operations end-to-end (with ordered hydration/stream tests), or remove the dead wire implementation and narrow the PR claims before merge.

shared/rust-bridge/codex-mobile-client/src/mobile_client/mod.rs
```diff
@@ -1754,23 +1841,42 @@ impl MobileClient {
...
                     presentation: None,
                     capabilities: None,
+                    local_studio: None,
                 },
             )]
         } else {
             requested_agents
+                .into_iter()
+                .filter(|(runtime_kind, agent)| {
+                    !is_local_studio_controller_agent(runtime_kind, agent)
```
0xSero: [P1] This does not create a distinct Local Studio Pi. It removes the controller-only `local-studio` advertisement and chats through the ordinary `pi` stream. The pinned Alleycat branch constructs exactly one `PiBridge`; when a Local Studio `pi-agent/models.json` exists it globally sets `PI_CODING_AGENT_DIR` on that one bridge. Therefore generic KittyLitter pairing and Local Studio Profile pairing on the same host address the same Pi identity/catalog, so the machine Pi and Local Studio Pi cannot coexist as required. Please have Alleycat advertise a separate Local Studio Pi chat runtime/resource (leaving `local-studio` as controller JSON-RPC), map that typed identity on mobile, and add an acceptance test proving both Pi catalogs remain isolated on one host.

services/kittylitter/Cargo.toml
```diff
@@ -22,9 +22,8 @@ name = "kittylitter"
 path = "src/main.rs"
 
 [dependencies]
-# Tracks dnakov/alleycat main. `./tools/scripts/update-alleycat-main.sh`
-# refreshes Cargo.lock to whatever `main` points at before build/release lanes.
-alleycat = { git = "https://github.com/dnakov/alleycat.git", branch = "main" }
+# Keep the daemon and mobile client on the same signed Local Studio bridge wire.
+alleycat = { git = "https://github.com/dnakov/alleycat.git", branch = "codex/litter-local-bridge" }
```
0xSero: [P1] Release behavior depends on an unreviewed mutable branch outside this PR. `codex/litter-local-bridge` currently resolves to `b855393…`, but there is no PR for that branch in `dnakov/alleycat`; moreover `tools/scripts/update-alleycat-main.sh` is invoked by build/release lanes and rewrites the lock to the branch head. A later push or branch deletion can change or break artifacts without changing this Litter PR SHA, and the core grant/Pi behavior is not reviewable here. Land and review the Alleycat changes first, then consume an immutable revision or reviewed main commit; add a reproducibility check that release builds do not mutate the lockfile.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
```diff
@@ -1104,6 +1219,31 @@ impl AppClient {
                     );
                 }
             }
+            if server_id.starts_with("alleycat:local-studio:") {
```
0xSero: [P2][Pi review] SSH Local Studio bypasses controller-model scoping. Both platform SSH pickers expose the `local-studio` runtime and the SSH bridge launches Pi with the Local Studio agent directory, but this guard only scopes models for `alleycat:local-studio:*`. An SSH Local Studio server therefore receives the full Pi catalog instead of only that controller’s active models. Carry a typed Local Studio server identity through SSH connections and scope it identically (or remove that SSH option until supported). Add an SSH Local Studio model-list test that proves unrelated provider models cannot appear.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
```diff
@@ -1104,6 +1219,31 @@ impl AppClient {
...
+                            error.error.message
+                        );
+                        Vec::new()
+                    }
+                    Err(error) => {
+                        tracing::warn!("Local Studio model scope failed: {error}");
+                        Vec::new()
+                    }
+                };
+                retain_local_studio_controller_models(&mut models, &active_model_ids);
```
0xSero: [P2][Pi review] A controller lookup failure is converted into an empty active-model list and then used as authoritative scope, stripping every Pi model even when Pi `model/list` succeeded. This recreates the observed empty/loading selector during a transient dashboard/controller failure. Preserve the last successfully scoped catalog and do not overwrite it when controller scope cannot be loaded; add a regression test covering a successful Pi list followed by a controller transport/error result.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
```diff
@@ -1104,6 +1219,31 @@ impl AppClient {
...
             }
+            if server_id.starts_with("alleycat:local-studio:") {
+                let active_model_ids = match c.load_local_studio_controller(&server_id).await {
+                    Ok(crate::local_studio::LocalStudioControllerLoadResult::Loaded {
+                        snapshot,
+                        ..
+                    }) => snapshot
+                        .sections
+                        .status
+                        .value
```
0xSero: [P2][Claude review] The outer controller call can succeed while `snapshot.sections.status.value` is `None` because status has independent error/freshness state. `unwrap_or_default()` then treats a transient section failure as an authoritative empty allow-list and strips every Pi/Local Studio model. Preserve the last successfully scoped catalog when the status section has no value, just as for the outer error arms. Add a test for successful Pi `model/list` plus a loaded controller snapshot whose status section failed.

0xSero:
Independent review record for `baecfa7` (draft — do not merge yet)

Reviewers:
- Claude Sonnet, through OMP’s authenticated provider because the direct Claude Code OAuth session was expired: corroborated the fresh-grant, distinct-Pi, dead session/turn, mutable-dependency, and SSH-scoping concerns; also found the missing-status catalog wipe.
- OMP / `openai-codex/gpt-5.4`: independently corroborated distinct Pi, dead session/turn paths, SSH scoping, missing-status catalog wipe, and mutable dependency.
- Pi / `openai-codex/gpt-5.4`: independently found the Local Studio permission presentation issue, SSH scoping, and controller-failure catalog wipe.
- Primary review left the evidence-backed inline discussions on each merge blocker and reviewer finding.

Contained fixes added after review:
- Register a fresh Local Studio pairing target before attachment.
- Keep the scoped model catalog when controller status is missing or temporarily unavailable.
- Treat Local Studio Pi as fixed full access on iOS and Android.
- Prevent controller-only Local Studio catalogs from being replaced by the generic Pi catalog.
- Key the iOS pairing sheet by pairing mode so Local Studio cannot reuse the generic kittylitter scanner.
- Make both “Scan Pairing QR” and “Paste Pairing JSON” directly reachable, and stop the camera off the main thread.

Remaining merge blockers:
1. Fresh Profile QR pairing still does not itself grant the required `stats.read` capability.
2. Local Studio does not yet advertise a truly distinct Pi runtime; the current upstream bridge overlays one global Pi.
3. Session-read and agent-turn code remains private/unwired and has no end-to-end chat surface.
4. SSH-selected Local Studio does not yet enforce controller-scoped model discovery.
5. Release builds still resolve a mutable, unreviewed Alleycat branch.

Validation:
- Android debug Kotlin compile: passed.
- Android debug unit tests: passed.
- Shared Rust targeted Local Studio contract test: 1 passed.
- iOS simulator build: passed.
- Reinstalled and exercised the Local Studio chooser, Profile QR instructions, and JSON fallback in the iPhone 17 Pro simulator.
- Gitleaks scanned the full two-commit PR range: no leaks found.
- Final PR size: 4,636 additions + 353 deletions = 4,989 changed lines.
- New CI for this SHA is queued.

Relevant discussions:
- Fresh grant: https://github.com/0xSero/litter/pull/186#discussion_r3634405611
- Distinct Pi: https://github.com/0xSero/litter/pull/186#discussion_r3634405851
- Unwired session/turn: https://github.com/0xSero/litter/pull/186#discussion_r3634405842
- Mutable dependency: https://github.com/0xSero/litter/pull/186#discussion_r3634408593
- SSH scoping: https://github.com/0xSero/litter/pull/186#discussion_r3634456099

0xSero:
Final iOS new-session latency fix is at `1a2c5c1`: the Home composer now opens the new conversation as soon as `thread/start` returns, before recent-directory bookkeeping, the first `turn/start` continues without blocking navigation, and the redundant post-start resume/list retry loop is removed. Verified on the freshly installed iPhone 17 Pro simulator with a live MacBook Local Studio Pi send; the conversation opened and rendered the first user message within the 5-second smoke assertion. `make ios-sim-fast`, incremental `xcodebuild`, `git diff --check`, and staged Gitleaks all passed. Final PR range is exactly 5,000 changed lines.

0xSero:
## Deep review — Local Studio integration across `litter`, `alleycat`, and `local-studio`

I read all three codebases end to end (shared Rust, iOS, Android, `alleycat` @ `codex/litter-local-bridge`, and `sybil-solutions/local-studio` @ `2.1.0`), probed the live daemon and the live Local Studio gateway on this machine, and diffed against the abandoned `docs/local-studio-integration-scope` branch.

The PR itself is reasonable. The **ecosystem around it is not**, and two findings below are release blockers that are not in the existing blocker list.

---

### 1. `kittylitter local-studio revoke` no longer does anything

This is the most serious finding and it is user-facing security.

`crates/alleycat/src/agents.rs:494`:

```rust
"local-studio" => {
    let _ = authenticated_node;          // <-- caller identity discarded
    let bridge = self.local_studio_bridge.clone()
        .ok_or_else(|| anyhow!("Local Studio runtime is unavailable"))?;
```

and the gate is now membership-free:

```rust
pub fn local_studio_gateway_available(&self) -> bool {
    self.local_studio_bridge.is_some()
}
```

`cargo check -p alleycat` confirms the whole capability layer is unreachable:

```
local_studio.rs:93:15:  warning: function `local_studio_agent_info` is never used
local_studio.rs:144:4:  warning: function `authorized_gateway` is never used
local_studio.rs:42:7:   warning: constant `IMPLEMENTED_CAPABILITIES` is never used
local_studio.rs:483:19: warning: struct `LocalStudioBridge` is never constructed
```

`local_studio_agent_info` was the only reader of `GrantStore::effective`. Meanwhile `kittylitter local-studio list|grant|revoke` is still a shipping CLI that writes `paired-nodes.json` and prints grants. On this machine right now:

```
$ kittylitter local-studio list
8032ef96…  grants=stats.read,sessions.read,agent.turn  expires=2026-07-28T00:00:00+00:00  revoked=no
88f77a2c…  grants=stats.read,sessions.read,agent.turn  expires=2026-07-28T00:00:00+00:00  revoked=no
ea4a6c63…  grants=stats.read,sessions.read,agent.turn  expires=2026-07-28T00:00:00+00:00  revoked=no
```

Those grants govern nothing. **A user who runs `revoke` to cut off a lost phone will believe they succeeded and will not have.** Any holder of the host pair token can `Connect{agent:"local-studio"}` and get the full app-server surface — `turn/start`, `command/exec`, `thread/read` — over Local Studio's Pi home, with `sandbox: danger-full-access` forced on (`mobile_client/mod.rs:1063`).

Note the grants above also **expire 2026-07-28**, four days out. Under the old design every paired device would have silently lost access on that date. That is presumably what the uncommitted v1.6.0 `pairingWipeGeneration` migration is reacting to.

Either delete the grant CLI and the ~5,000 lines of dead Design-A code, or re-wire the check. Shipping a revocation control that no-ops is worse than shipping neither. Tracked in the alleycat repo.

---

### 2. Three complete implementations exist; the one that ships is the smallest

| Where | What | Lines | State |
|---|---|---|---|
| `local-studio` `services/agent-runtime/src/litter-bridge-gateway.ts` + `shared/agent/litter-bridge.ts` + mutation ledger | Signed, versioned HTTP bridge: Ed25519 per-request auth, canonical JSON body hashing, replay store, idempotency ledger, optimistic `revision` concurrency, 16 typed error codes, 26 security tests | ~4,600 | **live and running right now** on `127.0.0.1:8081`, used by nobody |
| `alleycat` `crates/local-studio-proto` + `crates/alleycat/src/local_studio.rs` | The Rust half of the same protocol, 40 passing tests incl. a cross-impl golden signature vector | ~5,000 | **compiled, tested, dead** |
| `litter` `docs/local-studio-integration-scope` | Wiki, 3 ADRs, threat model, test/acceptance matrix, ops runbook | ~6,400 | abandoned on a branch |
| **this PR** | shell-script path probe + `ProcessLauncher` decorator | **198** | ships |

I verified the gateway is live:

```
$ curl -s -XPOST -H "content-type: application/json" \
    -H "x-local-studio-litter-bridge-secret: $SECRET" \
    http://127.0.0.1:8081/api/litter-bridge/v1 -d '{}'
{"type":"error","protocolVersion":1,"requestId":"fde0628f-…",
 "error":{"code":"unsupported_version","message":"Gateway request is invalid","retriable":false,…}}
```

Local Studio's contract already defines exactly the thing this integration is missing:

```ts
export const LitterBridgeSessionAuthoritySchema = Schema.Literals(["local-studio", "litter"]);
// + SessionTransferEnvelope, TransferCursor, TransferAck, ContentHashes,
//   ExternalSessionIdentity, SessionCreateRequest
```

That is bidirectional session handoff — start on desktop, resume on phone and back — designed, schema'd, and partly implemented. **This PR cannot do that**, because Design B (Local Studio as a second `PiBridge`) has no session-authority concept at all.

**I am not arguing to revert to Design A.** Design B is the right call for chat: it reuses the entire Pi runtime, thread list, model picker, and tool timeline for ~198 lines. That is genuinely good engineering. The problem is that Design A was left *in place* rather than deleted or wired, in two repos, with its security CLI still exposed. Pick one and delete the other.

---

### 3. Verified defects in this PR

**a. iOS — a keychain failure now silently discards a successful pairing.** `AlleycatAddServerSheet.swift:472`

```swift
let result = try await appModel.serverBridge.connectRemoteOverAlleycat(...)   // already succeeded
try AlleycatCredentialStore.shared.saveToken(params.token, nodeId: params.nodeId)
```

`main` had this non-fatal:

```swift
do { try AlleycatCredentialStore.shared.saveToken(...) }
catch { NSLog("[ALLEYCAT_CREDENTIALS] keychain save failed: %@", error.localizedDescription) }
```

Now a throw skips `onConnected(...)`, so `SavedServerStore.rememberAlleycat` never runs. Rust is connected, the user sees a red error, and the server is gone on next launch. Regression vs `main`.

**b. Android P0 — the iroh device key is never persisted at pair time.** iOS calls `AppRuntimeController.shared.persistAlleycatSecretKeyIfNeeded()` right after `saveToken`. Android's `connect()` (`AlleycatAddServerSheet.kt:233`) goes straight from `credentialStore.saveToken(...)` to `onConnected(...)`. `AppModel.persistAlleycatSecretKeyIfNeeded()` exists but is only called from `reconnectSavedServers()` and `onResume()`. Pair → kill the app without a background cycle → new `EndpointId` next launch → the host no longer recognises the device. This is precisely the bug `0a7143b` fixed on iOS.

**c. Android P1 — blank agent list.** `AlleycatAddServerSheet.kt:452` guards on `agents.isEmpty()` but iterates `availableAgents`. iOS guards on `availableAgents.isEmpty`. When every probed agent is `available == false` — the likely Local Studio first-run state — Android renders an empty box, no message, dead Connect button.

**d. `resolve_runtime` doesn't source the login profile.** `ssh_bridge.rs:240` builds the probe as `"{PROFILE_INIT}\n{...}"`; `local_studio.rs:85` builds the resolve as `format!("{FIND_AGENT_DIR}\n{FIND_RUNTIME}\nfind_local_studio_runtime")` with no `PROFILE_INIT`. The PATH fallback tier (`command -v pi-coding-agent || command -v pi`) therefore resolves during probe and fails during connect — the picker shows Local Studio **Available**, then connect fails with "Local Studio controller runtime was not found on the SSH host."

**e. Android P1 — camera-denied is a dead end.** iOS offers an "Open Settings" alert; Android shows static text with no `Settings.ACTION_APPLICATION_DETAILS_SETTINGS` intent. On Android a second denial is permanently sticky.

**f. `refresh_models` changed its contract.** It now returns `Err(ClientError::Rpc(failures.join("; ")))` on *partial* failure, after already committing the successful catalogs. Every refresh with Local Studio offline will surface an error banner even though the model list is fine.

---

### 4. CI runs zero tests

`.github/workflows/mobile-ci.yml`: the `android` job ends at `:app:assembleDebug`, the `ios` job ends at `xcodebuild … build`, and `grep -rn "cargo test" .github/workflows/` returns nothing. "GitHub Mobile CI: shared prep, Android, and iOS all passed" means **it compiles**. The 777 Rust tests in the PR description were run locally and have never run in CI.

This matters more than usual here because there is **no Android toolchain on the development machine** — no `~/Library/Android/sdk`, no NDK, no AVD, no `adb`. Android literally cannot be exercised locally, so CI is the only possible source of Android signal, and it currently provides none. That is the mechanical reason this feature has never been run on Android.

Feature test coverage today: iOS 0 of 29 test files, Android 0 of 13, shared Rust 3 (`parse_runtime` ×2, `local_studio_thread_start_is_always_full_access`). `RuntimeLauncher::launch`, `probe_script()`, and `resolve_runtime` — the entire integration — are untested.

---

### 5. Policy duplicated across three languages

`runtime == "pi" || runtime == "local-studio"` is written **six** times:

- `mobile_client/mod.rs:1063`, `mod.rs:1719`, `ffi/client.rs:161` (Rust ×3)
- `AgentRuntimePresentation.swift:99` (Swift)
- `AgentRuntimePresentation.kt:89` (Kotlin)
- plus `ffi/client.rs:206` with *different* membership (`"pi" | "local-studio" | "opencode"`)

`AppAgentCapabilities` already crosses UniFFI and already carries `supports_thread_permission_overrides`. This should be one `has_fixed_full_access` flag set by Rust. Same class: `alleycat:local-studio:<nodeId>` is minted by Swift string interpolation with no Rust constructor or parser; `AppAlleycatAgentWire → String` is hand-switched in three Swift sites and re-parsed in Rust as `== "ssh-bridge"`.

`local_studio::RUNTIME_KIND` exists but 7 call sites use the bare literal.

---

### 6. Answers to the six questions

1. **Is this too much?** No — the PR is too *little*, and the ecosystem is too much. 198 lines of glue is proportionate. ~23,000 lines of parallel abandoned implementation across three repos is not.
2. **Anything already written we failed to use?** Yes, extensively — see §2. The most painful: Local Studio's `litter-bridge/v1` gateway is running on this machine right now and this PR reads its `litter-bridge.json` **only to `sed` out a `pid`**, ignoring `url`, `secret`, `secretHeader`, `controllerId`, and `protocolVersion`. And that pid path is `/proc`-based, so it never fires on macOS.
3. **Can it survive production?** Not as-is. Blockers: the no-op revocation (§1), the iOS pairing-loss regression (§3a), the Android P0 (§3b), and zero test signal (§4). The `alleycat` pin is also on an unmerged feature branch that has already moved two commits past the pinned rev — `ad5dba33` vs branch head `b313786`.
4. **What to improve?** §3 defects, §5 de-duplication, §4 CI, and a decision on §1/§2.
5. **Docs / shared contract?** A good contract exists (`shared/agent/litter-bridge.ts`, protocol v1, strict, test-vector-pinned) and a good wiki exists (`docs/local-studio-integration-scope`). Neither is referenced from anywhere either team would look. `alleycat` appears **zero** times in the local-studio repo; there is no `docs/litter-bridge.md`.
6. **Performance measurement?** The hooks already exist and are unused: every gateway response carries `requestId`, `revision`, and per-section `Freshness {observed_at, age_ms, max_age_ms, stale}`. `kittylitter probe --resume-from/--repeat-resume-from` reproduces reconnect/replay exactly like the phone. `refresh_models` now fans out with a 20s per-runtime timeout but records no timing. Concrete proposal in the follow-up issue.

---

### Plan

Working through these in waves; PRs will follow. Wave 1 is the six verified defects in §3 plus CI test execution, because those are small, provable, and independently valuable regardless of how the §1/§2 architectural question is resolved.

Cross-repo items are filed separately so they do not block this PR.

0xSero:
## Release readiness — KittyLitter 0.3.x and mobile 1.6.0

Follow-up to the review above, covering what a release containing Local Studio would actually consist of and what currently blocks it.

### What KittyLitter is, mechanically

`services/kittylitter` is a thin wrapper crate that re-exports the alleycat daemon under a KittyLitter-branded name. cargo-dist 0.31 builds it for five targets, macOS binaries are codesigned with the Developer ID cert, and **npm is the only distribution channel** — Homebrew, shell/PowerShell installers, and MSI are all configured-but-disabled in `dist-workspace.toml`. Publishing uses npm trusted publishing via OIDC, so no `NPM_TOKEN`.

That is a clean setup. The problem is not the pipeline, it is what the pipeline compiles.

### Blocker 1 — the release compiles unreviewed code from a moving branch

`services/kittylitter/Cargo.toml` and `shared/rust-bridge/Cargo.toml` pin all five alleycat crates to `rev = "ad5dba33…"` on `codex/litter-local-bridge`, a branch in `dnakov/alleycat` with **no PR**. Since that pin was set the branch has moved four commits:

```
ad5dba3 (pinned) → 0f78f3b → b313786 → 84cb342
```

So the daemon we would ship is built from a feature branch that nobody has reviewed, that keeps moving, and whose contents are not visible in this PR. This was raised as a P1 earlier in the review and is still open.

The `update-alleycat-main.sh` change in this PR does fix the *lockfile-mutation* half — it now skips refresh when a `rev` is present. The remaining half is that the rev itself is not a reviewed commit. Tracked at `0xSero/alleycat#31` and `0xSero/alleycat#32`; the fix is to land that branch into `main` and pin a `main` commit.

### Blocker 2 — the grants in the field expire 2026-07-28

Every Local Studio grant on this machine carries `expires=2026-07-28T00:00:00+00:00`. Under the design those grants came from, every currently-paired device would lose access on that date with no in-app renewal path. That is three days out.

This is now moot *only because* grants no longer gate anything (`0xSero/alleycat#31`) — which is not a reassuring reason for it to be fine. Whichever way #31 is resolved, expiry-and-renewal needs an answer before a release goes out, not after.

### Blocker 3 — the 1.6.0 bump ships a destructive migration

The stashed working-tree changes bump iOS to `1.6.0 (2)` and Android to `versionCode 12 / 1.6.0`, and add a `pairingWipeGeneration = 2` migration that, on first launch, wipes **every** Alleycat token, the iroh device secret key, and every Alleycat-paired saved server. Every existing user re-pairs by scanning a QR.

That may well be the intent — it is the clean way out of the expiry problem above. But it should be a deliberate, release-noted decision, not something that rides along in a stash. Two specific things to fix before it ships:

- iOS stamps the generation only after `wipeAll()` succeeds; Android uses `prefs.edit().apply()`, which is fire-and-forget, so `runCatching` can never observe a failure and the generation is stamped regardless. Android should use `.commit()` to match.
- iOS runs the wipe in `AppRuntimeController.bind()`; Android runs it in the `AppModel` constructor, i.e. synchronous keychain/disk I/O inside `MainActivity.onCreate`. Move it off the cold-start path.

### Blocker 4 — no release lane runs tests

`release.yml`, `mobile-release.yml`, `ios-testflight.yml`, and `android-play-release.yml` all build. None run a test suite, and until #188 neither did `mobile-ci.yml`. A tagged release today would be gated on nothing but compilation.

`0xSero/litter#188` adds `cargo test`, `:app:testDebugUnitTest`, and `xcodebuild test` to CI. The release lanes should depend on that same signal rather than duplicating it.

### What a Local Studio release should contain

In dependency order:

1. `0xSero/alleycat#32` merges, then `codex/litter-local-bridge` lands in `dnakov/alleycat` `main` — resolving #31 either way.
2. Litter re-pins both manifests to a reviewed `main` commit and regenerates both lockfiles.
3. `0xSero/litter#188` merges (CI runs tests; pairing defects fixed).
4. `#186` merges.
5. The 1.6.0 bump lands as its own commit with the migration fixes above and an explicit "all devices must re-pair" release note.
6. Tag → cargo-dist publishes kittylitter to npm → TestFlight and Play internal.

### Keeping it scalable and publishing everywhere

Three things would remove most of the recurring cost:

**Make the alleycat dependency a normal dependency.** Every release problem above traces back to consuming alleycat as a git dep on a branch. Publishing the five crates to crates.io — or at minimum always pinning to `main` commits with a scheduled bump PR — turns a release-time hazard into routine dependency maintenance.

**Turn the E2E harness into a release gate.** `tools/scripts/local-studio-e2e-proof.sh` (added in #188) already asserts the eight things that actually matter: availability, catalog, start-and-complete-a-turn, reconnect-from-cursor, both handoff directions, and filesystem access. It needs a live daemon so it cannot run on hosted CI, but it can run on a self-hosted runner or as a documented pre-tag manual gate. Right now the release checklist has no behavioral gate at all.

**Re-enable the disabled installers once the dependency is stable.** Homebrew, shell, and PowerShell installers plus MSI are already configured in `dist-workspace.toml` and commented out. The reason to keep them off is the churn above, not any technical obstacle — `aarch64-pc-windows-msvc` is the one genuine gap and needs cargo-xwin/ring coverage. "Publish everywhere" is mostly a matter of making releases boring enough to trust.

0xSero:
The performance question from the review is now concrete at #189 — the short version is that `_alleycat_seq`, `SessionInfo{attached,current_seq,floor_seq}`, and `kittylitter probe` already give us everything needed to measure connect, model-list fan-out, new-session handoff, and time-to-first-token; none of it is recorded today, which is why the latency work in `6666098`/`1a2c5c1` had to be validated by stopwatch.

0xSero:
## Re-review after folding in #188

#188 is rebased in (8 commits) and closed. This PR is now **+2,589 / −573 across 66 files** and self-contained. Re-reviewing the whole thing against the five original merge blockers plus what I found in the audit.

---

## The five original blockers: 4 are obsolete, 1 stands

**1. "Fresh Profile QR pairing does not grant `stats.read`" — obsolete.**
Correct at the time. It no longer applies because the grant model gates nothing at all: `agents.rs` dispatches `local-studio` with `let _ = authenticated_node;` and `cargo check` reports `local_studio_agent_info` — the only reader of `GrantStore::effective` — as dead code. There is no capability to fail to grant. This became a *worse* problem in a different place, tracked at `0xSero/alleycat#31`.

**2. "Session read/turn is private, unwired dead code" — resolved by deletion.**
`ee9a35f` removed it. `local_studio.rs` is now 198 lines of SSH runtime discovery with no session or turn code, and no dead wire implementation. The PR no longer claims those capabilities.

**3. "Does not create a distinct Local Studio Pi" — resolved, in alleycat.**
This was the substantive blocker and it was fixed properly upstream, not papered over. `f443a83` stopped redirecting the user's `pi` bridge and built a second `PiBridge`; `53fb190` isolated its thread index under `<agent_dir>/bridge-index`; `01643f0` added `model_provider_prefix("local-studio")` so `model/list` is scoped server-side. Machine Pi and Local Studio Pi now coexist on one host with separate catalogs, sessions, and indexes.

**4. "SSH Local Studio bypasses controller-model scoping" — obsolete.**
The client-side scoping guard it referenced is gone; `grep alleycat:local-studio ffi/client.rs` returns nothing. Scoping moved into alleycat's `model_provider_prefix`, which is the right layer — one enforcement point that both the paired and SSH paths inherit.

**5. Two P2s about the controller snapshot wiping the catalog — obsolete.**
Same removal. `controller_snapshot`, `sections.status`, and the `unwrap_or_default()` those findings pointed at no longer exist in mobile.

**6. "Release depends on an unreviewed mutable branch" — STILL OPEN, and worse.**

This is now the only original blocker left, and it is the one thing I would not merge without.

`shared/rust-bridge/Cargo.toml` and `services/kittylitter/Cargo.toml` pin `rev = "ad5dba33…"` on `codex/litter-local-bridge` in `dnakov/alleycat` — a branch that **still has no PR**. It has moved five commits past the pin:

```
ad5dba3 (pinned) → 0f78f3b → b313786 → 84cb342 → … → 2353933
```

The `update-alleycat-main.sh` change in this PR correctly fixes the lockfile-mutation half. The remaining half is that the pinned commit is unreviewed code on a branch that can move or disappear.

**Blocking action:** open a PR for `codex/litter-local-bridge` → alleycat `main`, land it, re-pin both manifests to a `main` commit, regenerate both lockfiles.

---

## What #188 fixed

All verified against `main`, all now in this branch:

- iOS: a keychain failure silently discarded a successful pairing (`main` had a non-fatal `do/catch`; this branch had made it a bare `try`)
- Android P0: the iroh device key was never persisted at pair time — pair then kill regenerated the `EndpointId` and the host stopped recognising the device
- Android P1: the empty-agent state was unreachable (`agents.isEmpty()` guard, `availableAgents` body)
- Android P1: permanently-denied camera was a dead end
- Shared Rust: `resolve_runtime` did not source the login profile that `probe_remote_agents` uses, so Local Studio showed **Available** then failed to launch
- 7 new tests for `RuntimeLauncher`, which had none
- CI now runs tests in all three jobs — it previously ran zero
- ADR-0004, and a QA matrix row that had been describing a dashboard deleted in `ee9a35f`

---

## Still open — none blocking except #6 above

**`refresh_models` returns `Err` on partial failure.** `ffi/client.rs:1146` commits the successful catalogs via `update_server_models` and *then* returns `Err(ClientError::Rpc(failures.join("; ")))`. Every refresh with Local Studio offline surfaces an error banner over a model list that is actually fine. Worth deciding deliberately: either return `Ok` with a per-runtime error map, or have callers treat a throw as advisory.

**Local Studio wins reconnect auto-selection.** `reconnect.rs:834` puts `"local-studio"` first in the fallback candidate list, ahead of `claude`/`pi`/`codex`. A saved server with no explicit runtime now prefers Local Studio. Probably intended; flagging because it silently changes behaviour for existing servers.

**`runtime == "pi" || runtime == "local-studio"` is written six times** — Rust ×3 (`mod.rs:1063`, `mod.rs:1719`, `ffi/client.rs:161`), Swift, Kotlin, plus a fourth Rust variant with different membership at `ffi/client.rs:206`. `AppAgentCapabilities` already crosses UniFFI and already carries `supports_thread_permission_overrides`; this wants one `has_fixed_full_access` flag. Deliberately not in #188 — it touches many files and deserves its own diff.

**Two iOS suites are skipped in CI** (`HomeDashboardSupportTests`, `WatchCompanionBridgeTests`) because they read ambient state and go green locally while failing on a clean runner. 166 tests gate. Root causes and fixes in #190.

**Android has still never run on a device or emulator.** There is no Android SDK/NDK/AVD on the development machine, so CI is the only signal and CI is compile + unit tests. The pairing sheet, the QR scanner inside a `ModalBottomSheet`, and the new settings-intent path are unexercised by anything.

**The E2E harness is not a gate.** `tools/scripts/local-studio-e2e-proof.sh` asserts the eight things that matter — availability, catalog, start-and-complete-a-turn, reconnect-from-cursor, both handoff directions, filesystem access — but needs a live daemon, so it cannot run on hosted CI. It should be a documented pre-tag manual step or a self-hosted job.

---

## Verdict

The architecture question that dominated the original review is settled and settled correctly: Local Studio is a scoped Pi runtime, the isolation is enforced in alleycat where it belongs, and the mobile side is ~200 lines of glue rather than a second protocol stack. ADR-0004 records that and supersedes the three ADRs on the abandoned docs branch.

**Merge once the alleycat pin points at a reviewed `main` commit.** Everything else on this list is a follow-up, not a blocker.

CI: `shared-prep` 773 passed, `android` assemble + unit tests, `ios` 166 tests — re-running now against the rebased branch; I'll confirm.

0xSero:
## Slim-down review: ~5,900 deletable lines, and one cross-repo change that removes the brittlest code

Measured on the merged branch. Ordered by value.

---

### 1. alleycat — delete Design A entirely: **~5,945 lines**

| file | lines | live? |
|---|---:|---|
| `crates/local-studio-proto/src/lib.rs` | 2,450 | only `LocalStudioAdvertisement`/`Capability` are referenced, and only by other dead code |
| `crates/alleycat/src/local_studio.rs` | 2,506 | **4 items live** — `PiRuntimeCommand`, `pi_agent_dir`, `bundled_pi_runtime`, and their path helpers. ~250 lines of 2,506 |
| `crates/alleycat/src/grants.rs` | 809 | store round-trips faithfully, gates nothing |
| `crates/alleycat/src/cli/local_studio.rs` | 180 | the `grant`/`revoke`/`list` CLI |

`cargo check -p alleycat` reports **37 dead-code warnings**, all from this. Deleting takes them to zero and lets clippy become blocking in CI.

Also dead: `AgentInfo.local_studio` — alleycat writes the advertisement, and litter's `AgentInfoWire` has no such field, so nothing reads it. And roughly **45 of ~50 Local Studio tests** exercise the dead path, which is why the live `LocalStudioLauncher` had zero coverage until `d59471e`.

Blocked on the `0xSero/alleycat#31` decision. If the answer is "delete", this is a single mechanical PR.

---

### 2. Both platforms — collapse the 6× duplicated policy predicate

`runtime == "pi" || runtime == "local-studio"` is written six times in three languages:

```
mobile_client/mod.rs:1063    let defaults_to_full_access = runtime_kind == "pi" || runtime_kind == "local-studio";
mobile_client/mod.rs:1719    matches!(runtime_kind, "pi" | "local-studio")
ffi/client.rs:161            matches!(runtime_kind.as_str(), "pi" | "local-studio")
ffi/client.rs:206            matches!(runtime_kind, "pi" | "local-studio" | "opencode")   ← different membership
AgentRuntimePresentation.swift:98    runtime == "pi" || runtime == "local-studio"
AgentRuntimePresentation.kt:88       this == "pi" || this == "local-studio"
```

The fourth one already disagrees with the other three, which is the drift arriving on schedule.

`AppAgentCapabilities` already crosses UniFFI and already carries `supports_thread_permission_overrides`. Adding `has_fixed_full_access: bool` deletes all six call sites' logic, removes the Swift `hasFixedFullAccess` static and its Kotlin twin, and makes the policy a host-advertised capability rather than a client-side string match — which is what `CLAUDE.md` asks for.

Net: roughly −40 lines, and one source of truth instead of six.

---

### 3. Cross-repo — the descriptor change deletes the brittlest code in the PR

This is the highest-leverage item and it isn't in this repo.

`local_studio.rs` embeds two shell functions (`FIND_AGENT_DIR`, `FIND_RUNTIME`, lines 11–56) that guess at Local Studio's install layout, including this:

```
"$app/Contents/Resources/app/frontend/.next/standalone/frontend/node_modules/@earendil-works/pi-coding-agent/dist/cli.js"
```

That path encodes Next.js standalone output *and* npm hoisting *and* the `@earendil-works/pi-coding-agent` package layout. `alleycat` carries an independent copy of the same guessing in `bundled_pi_runtime()`.

**This is not hypothetical — it broke on my machine today.** Two Local Studio versions were present (2.1.0 in `/Applications`, 2.3.0 running from a mounted DMG). `bundled_pi_runtime()` resolves `/Applications` by hardcoded path, so it launched the 2.1.0 Pi CLI against a data dir 2.3.0 owned. Result:

```
WARN alleycat_pi_bridge::pool::process: pi writer task: stdin write failed; exiting err=BrokenPipe
```

The Pi process died on startup, `thread/start` timed out after 45s with no response, and the E2E suite went from 8/8 to 6/8 — **with no error surfaced to the user**. The agent still advertised `available=true`, because availability is a filesystem check (`models.json` exists), not a liveness or version check.

If `sybil-solutions/local-studio#265` lands and the already-existing `litter-bridge.json` publishes `piAgentDir` and `piRuntime`, then:

- litter's `FIND_AGENT_DIR` + `FIND_RUNTIME` (~46 lines of shell) collapse to a JSON read
- alleycat's `bundled_pi_runtime()` + `runtime_from_app_bundle()` + `runtime_command()` (~90 lines) collapse the same way
- the version-skew failure above becomes impossible, because the descriptor is written by the *running* instance

Net: ~135 lines deleted across two repos, and the class of bug goes away rather than being papered over.

---

### 4. Smaller

**Availability should mean "will start", not "a file exists."** `pi_agent_dir()` treats `models.json` as the readiness marker. The runtime advertises as available with Local Studio not running and with a mismatched binary. Cheapest honest fix: have the descriptor carry the version, and refuse to advertise when the resolved runtime does not match it.

**One shell-script assembly point.** `local_studio.rs` splices `FIND_AGENT_DIR` + `FIND_RUNTIME` in two places (`probe_script()` at :60 and `resolve_runtime()` at :89) under different shell environments. That mismatch was the "shows Available then fails to launch" bug fixed in `1d7b548`. One `runtime_script()` helper used by both would make the divergence structurally impossible rather than relying on both call sites remembering `PROFILE_INIT`.

---

### Summary

| change | lines | blocked on |
|---|---:|---|
| Delete alleycat Design A | −5,945 | `alleycat#31` decision |
| `has_fixed_full_access` capability | −40 | nothing |
| Descriptor-driven runtime resolution | −135 | `local-studio#265` |
| Single shell-script assembly | −10 | nothing |

Two of these need no coordination and could go in now. The two large ones are both waiting on decisions rather than on work.

shared/rust-bridge/codex-mobile-client/src/local_studio.rs
0xSero: Addressed by the current architecture. The deleted private controller/stats protocol and stats.read grant are no longer in the pairing path. Local Studio Profile QR/JSON now pairs through the standard signed Alleycat endpoint and selects the distinct local-studio Pi runtime, so fresh enrollment does not depend on a pre-granted development endpoint.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
0xSero: Addressed. The dead custom session read/turn implementation was removed. Local Studio now uses the existing Pi app-server thread/list, thread/read/resume, turn/start, event reducer, and hydration paths end to end. The shared suite covering hydration, ordered item/tool boundaries, resume, steer, and reconnect passes (765 unit tests plus integration/doc tests).

shared/rust-bridge/codex-mobile-client/src/mobile_client/mod.rs
0xSero: Addressed. The host now advertises a separate local-studio runtime backed by its own PiBridge, Local Studio agent directory, sessions directory, bridge index, launcher environment, and local-studio model-provider scope. The ordinary machine Pi remains a separate pi runtime and catalog.

services/kittylitter/Cargo.toml
0xSero: Addressed for reproducibility. Both Alleycat consumers use immutable commit revisions, and update-alleycat-main.sh exits before cargo update when a revision pin is present. The release check left both lockfile hashes unchanged. The host behavior is inspectable at pinned commit 7aeb38ab0381a56b1bbd86231813efeca9d0d5a2 and cannot drift with a branch push.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
0xSero: Fixed in 42dde07. The SSH Local Studio PiBridge now applies model_provider_prefix(local-studio), matching the Alleycat Local Studio runtime and preventing unrelated Pi providers from entering the SSH model catalog.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
0xSero: Addressed. The obsolete client-side controller allow-list path was deleted; controller model scope is enforced by the Local Studio Pi bridge. Failed runtime refreshes preserve the last cached catalog for that runtime, and 42dde07 also prevents one failed runtime from turning a successful multi-runtime refresh into a blocking catalog error.

shared/rust-bridge/codex-mobile-client/src/ffi/client.rs
0xSero: Addressed with the same architecture change: the deleted snapshot status section is no longer used as an authoritative empty allow-list. Model scope comes from the Local Studio Pi provider/catalog, and transient per-runtime failures retain cached models instead of wiping the selector.

0xSero:
Final adversarial review complete at `42dde07`.

Fixed in the final pass:
- SSH Local Studio now applies the same `local-studio` provider scope as Alleycat, preventing unrelated Pi models from leaking into the selector.
- Multi-runtime model refresh preserves useful results and cached failed-runtime catalogs instead of surfacing a blocking error when another controller succeeds.
- Remote `thread/list` gets a realistic 10-second deadline instead of the regression-prone 3-second cutoff.
- iOS keeps its resume guard for the entire awaited resume lifecycle, preventing duplicate resume RPCs.
- Android pairing failures retain their exception context.

Independent review:
- Pi reviewed the complete working diff. Its verified duplicate-resume finding is fixed; the remaining reports were checked against current code and were non-issues or intentional bounded behavior.
- Claude Code was logged out. Two authenticated Claude/OMP fallback reviews were attempted, but both hit their deadlines without returning a verdict; no Claude pass is being claimed.
- Primary review found no remaining release-blocking defect after these fixes.

Verification:
- shared Rust: 765 passed, 0 failed, 5 live-environment tests ignored; all integration and doc tests passed
- Android: `:app:testDebugUnitTest` passed
- iOS: `make ios-sim-fast` passed twice, including after the resume-race fix
- kittylitter service build/test passed
- immutable Alleycat update proof left both lockfiles unchanged
- gitleaks: 19 commits scanned, no leaks
- final PR diff: 1,996 additions / 843 deletions

All seven inline review threads have been answered and resolved. CI for the pushed head remains the final remote gate.

0xSero:
Runtime-boundary verification at `8de062c`:

- Local Studio pairing now exposes/selects only the `local-studio` runtime on iOS and Android; the setup card no longer presents generic Pi as part of Local Studio.
- Live `model/list` against the running controller returned 24 `local-studio/*` models and no GPT models.
- Live `model/list` against Codex returned 7 GPT models (`gpt-5.6-sol`, `gpt-5.6-terra`, `gpt-5.6-luna`, `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex-spark`) and no Local Studio models.
- Live desktop -> Litter proof: a pre-existing Local Studio session was found by `thread/list` and read by stable thread id.
- Live Litter -> desktop proof: a Litter-created session was stored under `Local Studio Dev/pi-agent/sessions/...jsonl`, the same store the Electron app reads.
- Cursor reconnect and filesystem execution/search passed.
- iOS simulator build/install/launch passed; Android `:app:testDebugUnitTest` passed; focused shared Rust runtime/model tests passed; staged gitleaks found no leaks.
- PR size remains 1,999 additions / 846 deletions.

External runtime state found during the detached-turn scenario: Local Studio currently has `GLM-5.2-EXL3` running while the selected request was `GLM-5.2`, so the controller persisted and rehydrated the expected `model_not_running` 503. This is a selected-vs-running recipe mismatch, not lost session state.

Physical iPhone rebuild was attempted but this clean checkout has no Xcode account/provisioning profile for team `UH66Q8ZAYG`; simulator is updated, physical device is not falsely reported as updated.

0xSero:
**Correction (2026-07-27):** I re-verified N1 and N3 at the byte level and both were false positives — my mistake misreading `od -c`'s `\t` mnemonic (which denotes a TAB byte, 0x09) as the two characters backslash-t. Findings N1 and N3 below are struck; the valid findings (N2 and the reconnect ordering) are fixed in `73fedb2`.

- ~~N1~~ — `probe_script()`'s `${{runtime%%\t*}}` uses a **literal TAB byte** (0x09) in the pattern, confirmed via `perl` byte dump. It splits on the tab correctly. No bug.
- ~~N3~~ — `probe_script()` calls `find_local_studio_runtime`, which returns non-empty only when **both** the agent data dir and the Pi runtime binary resolve. `parse_agent_probe` marks `Available` iff that field is non-empty. So `Available` already means "dir + binary resolved," not "a file exists." The residual version-skew case (stale dir + a *global* `pi` on PATH launching against it) is the slim-down review's §3 descriptor-driven-resolution follow-up (`sybil-solutions/local-studio#265`), not a launch-failure bug.
- **N2 (fixed, `73fedb2`):** mobile bridge crates and the kittylitter daemon are now co-pinned to the same immutable commit `7aeb38ab` (`b8b0c113` was its ancestor).
- **Reconnect ordering (fixed, `73fedb2`):** `local-studio` moved from first to last in the empty-requested SSH-bridge fallback, restoring `main`'s `claude/pi/opencode/codex` precedence so existing saved servers keep their prior runtime. Still selectable, just not the silent default winner.

The 6× `runtime == "pi" || "local-studio"` predicate remains a documented follow-up, not a bug — the `ffi/client.rs:206` site I flagged as "divergent" is a *different* predicate (`runtime_kind_uses_qualified_catalog`, which correctly includes `opencode`), not drift. The proper fix is the UniFFI `has_fixed_full_access` capability flag already deferred in the slim-down review.

`cargo check` clean; `cargo test -p codex-mobile-client --lib`: 765 passed, 0 failed, 5 ignored.

---

<details><summary>Original (incorrect) findings, preserved for the record</summary>

Adversarial re-review against the current head (`8de062c`). I verified the "Addressed" replies against the code instead of trusting them, and looked for things the prior reviews missed.

## Verified fixed (claims hold)
- `resolve_runtime` prepends `PROFILE_INIT` (`local_studio.rs:85`). The §3d "Available then fails" cause from PROFILE_INIT is closed.
- iOS `saveToken` is back in `do/catch` and `persistAlleycatSecretKeyIfNeeded()` runs right after (`AlleycatAddServerSheet.swift:469-479`). The §3a pairing-loss regression is gone.
- Android persists the iroh device key at pair time (`AlleycatAddServerSheet.kt:237`) and offers `ACTION_APPLICATION_DETAILS_SETTINGS` on camera denial (`:658`). §3b/§3c/§3e closed.
- SSH Local Studio applies `model_provider_prefix("local-studio")` (`ssh_bridge.rs:555`), matching the paired path. §3f SSH-scoping blocker closed.
- `refresh_models` returns `Ok` unless every runtime failed, and reuses cached catalogs for failed runtimes (`ffi/client.rs:1095-1098`). The partial-failure banner regression is closed.
- Both alleycat deps are pinned to immutable `rev=` (no branch). The "mutable branch" blocker is closed in the lockfile-mutation sense.

## New findings the prior reviews missed

~~**N1 — `${runtime%%\t*}` in `probe_script()` is broken bash, not a tab split.**~~ *(false positive — see correction above)*

**N2 — The two alleycat consumers are pinned to two different immutable commits.** *(fixed in 73fedb2)*

~~**N3 — "Available" still means `models.json` exists, not "the runtime will start."**~~ *(false positive — see correction above)*

## Re-emphasized, not new
- `runtime == "pi" || runtime == "local-studio"` is still written 6× — documented follow-up, not a bug (see correction above).
- `reconnect.rs:836` puts `local-studio` first in the empty-requested fallback list — *(fixed in 73fedb2)*.

## Verdict
No new release blocker. Merge-ready.

</details>

Issue #180, opened 2026-07-13
iOS Model picker stays on “Loading models…” when Codex returns reasoningEffort: "ultra"


hordesnake1:
## Summary

Litter's model picker remains on `Loading models...` when a Codex app-server returns a model whose `reasoningEffort` is `"ultra"`.

This reproduces with Codex `0.144.3` and its GPT-5.6 model list: `model/list` includes `gpt-5.6-sol` with `reasoningEffort: "ultra"`. Litter v0.3.4 appears to map only through `Max`, so decoding/mapping this response fails and no model list reaches UI.

## Reproduction

1. Run Codex app-server from Codex `0.144.3`.
2. Connect Litter v0.3.4.
3. Ensure app-server `model/list` response contains a model with:
   ```json
   { "reasoningEffort": "ultra" }
   ```
4. Open model picker.

## Actual

Model picker stays on `Loading models...`.

## Expected

Litter accepts and renders every valid Codex `reasoningEffort` value, including `"ultra"`, or handles unknown future values without blocking the entire model list.

## Suggested fix

Add `ultra` to Litter's reasoning-effort enum/mapping (after `Max` if ordering matters). Prefer a forward-compatible fallback for unknown server values so one new enum value cannot make the picker permanently load.

## Environment

- Litter: v0.3.4
- Codex: 0.144.3
- Server transport: remote `codex app-server` over SSH tunnel
- A temporary WebSocket filter that removes only `reasoningEffort: "ultra"` from `model/list` responses makes the picker load, confirming the trigger.

jmilldotdev:
this needs to be fixed...

jmilldotdev:
https://github.com/0xSero/litter/pull/187

Shlok-Bhakta:
+1

0xSero:
Fixed on main by #193. `ReasoningEffort::Ultra` now crosses the Rust/UniFFI boundary, both Swift and Kotlin render/map Ultra, and the pinned Codex patch accepts max/ultra. Closing the concrete GPT-5.6 loading failure. A future unknown-enum compatibility policy can be tracked separately if needed.

PR #187, opened 2026-07-22
bridge: support ultra reasoning effort + unknown reasoning levels

jmilldotdev:
Solves https://github.com/0xSero/litter/issues/180

  - Recognizes ultra throughout the Rust bridge, iOS, and Android.
  - Makes pinned Codex accept max, ultra, and unknown future reasoning-effort values.
  - Degrades unknown values to Litter’s strongest known option (max) instead of failing the entire model/
    list response.
  - Leaves generated bindings untouched.
  - Adds focused assertions for eventual CI coverage


0xSero:
Thanks for the fix. I re-audited this against current main: Ultra now crosses the Rust/UniFFI boundary and is rendered by both Swift and Kotlin via merged #193, which also closed #180. This branch now conflicts in four shared-Rust files and no longer supplies the missing behavior it was opened for. Closing as superseded. A future unknown-enum compatibility policy should be proposed separately with a typed cross-platform contract rather than carried in this stale conflict.

PR #188, opened 2026-07-25
fix(local-studio): repair pairing defects, run tests in CI, add an E2E proof harness

0xSero:
Stacked on `codex/local-studio-integration` (#186). Wave 1 of the review in https://github.com/0xSero/litter/pull/186#issuecomment-5075623985: the defects that are small, provable, and worth fixing regardless of how the architectural question in that review is resolved.

**+314 / −8 across 5 files, plus 2 new files.**

## Defects fixed

**iOS — a keychain failure silently discarded a successful pairing.** `AlleycatAddServerSheet.swift:472`. `main` wrapped `saveToken` in a non-fatal `do/catch`; this branch made it a bare `try`. The `connectRemoteOverAlleycat` RPC has already succeeded at that point, so a throw skips `onConnected(...)` and `SavedServerStore.rememberAlleycat` never runs — Rust is connected, the user sees a red error, and the server is gone on the next launch. Restored.

**Android P0 — the iroh device secret key was never persisted at pair time.** iOS calls `persistAlleycatSecretKeyIfNeeded()` immediately after `saveToken`; Android's `connect()` went straight to `onConnected(...)`. The method exists on `AppModel` but was only reachable from `reconnectSavedServers()` and `onResume()`, neither of which fires while the app stays foregrounded after pairing. Pair → kill → the `EndpointId` is regenerated and the host no longer recognises the device. This is the bug `0a7143b` fixed on iOS.

**Android P1 — the empty-agent state was unreachable.** The guard tested `agents.isEmpty()` while the body iterated `availableAgents`. A host advertising only unavailable agents rendered an empty box, no message, dead Connect button — the likely Local Studio first-run state.

**Android P1 — permanently denied camera was a dead end.** A second denial is sticky on Android and the permission launcher stops prompting. Added a route into app settings next to the existing paste fallback.

**Shared Rust — Local Studio showed *Available*, then failed to launch.** `probe_remote_agents` builds its script as `"{PROFILE_INIT}\n…"` (`ssh_bridge.rs:240`); `resolve_runtime` did not. So the PATH fallback tier (`command -v pi-coding-agent || command -v pi`) could resolve during discovery and fail during connect, producing "Local Studio controller runtime was not found on the SSH host" after the picker had already shown it as available.

## Tests

`RuntimeLauncher` — the decorator that rewrites the process spec, i.e. the entire integration — had **zero** tests. Added seven covering program/arg rewrite, `Agent` vs `ToolCommand` roles, explicit `PI_CODING_AGENT_DIR` precedence, the no-prefix PATH fallback, `probe_script` output, `session_scan_prefix` quoting, and `parse_runtime` rejection.

Tests matching `local_studio` go from 6 to 13.

```
cargo test -p codex-mobile-client
test result: ok. 773 passed; 0 failed; 5 ignored
```

## CI now runs tests

Mobile CI ran **zero** tests: the android job ended at `:app:assembleDebug`, the ios job at `xcodebuild … build`, and no workflow invoked `cargo test`. "CI passed" meant the tree compiled.

This matters most for Android — there is no Android SDK, NDK, or emulator on the development machine, so this job is the only place Android behavior can ever be exercised. Added `cargo test` to `shared-prep`, `:app:testDebugUnitTest` to `android` (with the report uploaded on failure), and `build-for-testing` + `test-without-building` to `ios`.

## End-to-end proof

`tools/scripts/local-studio-e2e-proof.sh` drives `kittylitter probe`, which dials the daemon over iroh and speaks the same JSON-RPC a paired phone speaks — so these exercise the real transport, the real Local Studio Pi runtime, and the real session store, not a mock.

Run against the live daemon and the running Local Studio desktop app:

```
[1/7] agent availability
  PASS  availability     local-studio advertised as available
[2/7] controller-scoped model catalog
  PASS  catalog          model/list returns a catalog
[3/7] start session + complete a turn
  PASS  start            turn/completed observed after thread/start
        thread: 019f9691-b26e-7420-89dc-36214fe4e84f
[4/7] leave and return (reconnect from cursor)
        resuming from seq 11
  PASS  leave-return     reconnect with a cursor reattached the existing session
[5/7] resume a desktop-created session
        desktop thread: 019f9690-7510-7b63-8805-bbbfaed2bbb6
  PASS  resume-desktop   a pre-existing Local Studio session is readable over the phone path
[6/7] filesystem access
  PASS  fs-exec          shell command executed inside the session cwd
  PASS  fs-search        fuzzy file search returns repository paths
[7/7] phone-created session is visible to Local Studio
        path: /Users/…/Local Studio/pi-agent/sessions/--Users-sero-ai-projects-litter--/2026-07-25T00-26-07-296Z_019f96aa-….jsonl
  PASS  handoff-to-studio the phone-created session is stored in Local Studio's session store

passed 8, failed 0
```

Both handoff directions are covered: **desktop → phone** by scenario 5, and **phone → desktop** by scenario 7, which asserts the phone-created session lands in Local Studio's own `pi-agent/sessions/` store — the store the desktop UI reads — rather than somewhere only mobile can see.

The two scenarios that are easy to fake are asserted honestly:

- **leave/return** asserts on the *second* connect only, so a host that always minted a fresh session would fail. The captured evidence is `attached=Fresh current_seq=0` on the first connect and `attached=Resumed current_seq=2 resume_from=Some(11)` on the reconnect.
- **resume-desktop** requires a session whose `createdAt` predates the run *and* whose path is under Local Studio's own `pi-agent/sessions/` store, so the harness cannot pass by reading back its own output.

Evidence lands in `artifacts/local-studio-e2e/` (gitignored).

## Docs

**ADR-0004** records that Local Studio ships as a scoped Pi runtime and supersedes ADR-0001/0002/0003 on the unmerged `docs/local-studio-integration-scope` branch, which specified a signed capability-scoped controller plane. That branch's own rule is that changes to state ownership, trust boundaries, or session identity require a new ADR. Both halves of the superseded design still exist — live and unconsumed in local-studio, compiled and dead in alleycat — so the decision needed writing down rather than inferring. The ADR names the accepted risks (full-access trust level, discovery brittleness, dead code in two repos) and links the cross-repo follow-ups.

**QA matrix**: the Android Local Studio row described a Settings controller dashboard **deleted in `ee9a35f`**, described an SSH-first flow that is not the shipped one, and claimed `Same` parity for a feature no Android device has run. Replaced with the actual flow and an explicit untested marker.

## Not in this PR

- The `alleycat` grant model being a no-op — `0xSero/alleycat#31`.
- Local Studio publishing a runtime descriptor instead of forcing path guessing — `sybil-solutions/local-studio#265`.
- De-duplicating `runtime == "pi" || runtime == "local-studio"` across Rust ×3, Swift, and Kotlin into an `AppAgentCapabilities` flag. Wave 2 — it touches many files and deserves its own review.
- The uncommitted v1.6.0 version bump and `pairingWipeGeneration` migration, which were already excluded from #186 and are stashed locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


0xSero:
## CI results — and what turning tests on immediately found

**`shared-prep: pass`** — the shared Rust suite ran in CI for the first time:

```
test local_studio::tests::agent_launch_uses_the_bundled_electron_runtime ... ok
test result: ok. 773 passed; 0 failed; 5 ignored
```

**`android: pass`** — `:app:testDebugUnitTest` green.

**`ios: fail`** — and this is the useful part. `build-for-testing` succeeded, so the test targets compile fine; the *test run* failed with **9 pre-existing failures**. None are caused by this branch or #186 — `AppThreadSnapshot+Runtime.swift`, `SavedAppsStore.swift` and `WatchCompanionBridge.swift` are all unchanged since `main`. They were invisible because nothing had ever run them.

### The suite is not hermetic

The same assertion fails **differently** depending on the machine. On GitHub's runner:

```
testHomeDashboardModelRefreshesWhenObservedSnapshotChanges:
  XCTAssertEqual failed: ("[]") is not equal to ("[\"server-a\"]")
```

Same test on my machine:

```
  XCTAssertEqual failed:
    ("[\"alleycat:local-studio:3e0bf5506496c8370908487dc6eaf448a6a0a4927a341f56e4f1f35cd9a02312\"]")
    is not equal to ("[\"server-a\"]")
```

It returned **my actual paired Local Studio server**. `HomeDashboardModel` is observing the real `AppModel`/`SavedServerStore` singletons instead of the injected snapshot, so the fixture is ignored either way. A green local run for that suite has been meaningless — it was asserting against whatever the developer happened to be paired to.

### What I did

Skipped the five known-bad tests with an inline comment and a tracking issue rather than fixing 9 unrelated failures inside a Local Studio PR. That leaves **191 tests gating**, verified locally:

```
Executed 191 tests, with 0 failures (0 unexpected)
```

Full per-test breakdown, root causes, and suggested fixes in #190. The real work there is making `HomeDashboardModel` take injected state instead of reaching for singletons; the rest are small. Each skip should come out as it is fixed.

I'd rather land a gate that is honest about what it does not yet cover than either hide the failures or balloon this PR.

0xSero:
## All checks green

| job | result | evidence |
|---|---|---|
| `shared-prep` | pass | `773 passed; 0 failed; 5 ignored` — shared Rust, first time ever in CI |
| `android` | pass | `assembleDebug` + `:app:testDebugUnitTest` both executed |
| `ios` | pass | `Executed 166 tests, with 0 failures` |
| `plan` | pass | |

### What it took

Three CI-only failures, none reproducible on a developer machine:

1. **`cloud_sync` race** (`ca3d7f0`) — `PLATFORM_TABLE` is a process-global mutex and 8 tests reset it while cargo runs them in parallel. Pass/fail was pure scheduling luck, which is why the first CI run went 773/773 and the next failed on one test. Now serialized behind a `#[must_use]` guard; verified 773/773 across 3 consecutive full-suite runs locally and again in CI.

2. **`WatchCompanionBridgeTests`** (`4ce78ef`) — skipping one WatchConnectivity test surfaced another. 12 of the suite's 26 tests assert pushes through a real `WCSession`, which a clean runner does not have. Skipped as a suite; per-test quarantine would have cost one CI round-trip per test.

3. **`HomeDashboardSupportTests`** — already suite-skipped in the first quarantine.

### The pattern worth keeping

Both skipped suites go **green on a developer machine and red on a clean runner**, because they read ambient state — saved servers, a real WCSession. That asymmetry is why they survived unnoticed: the only signal anyone had was a local run, and a local run was never a valid check for them.

166 tests gate today. The two skips come out together once `HomeDashboardModel` and the Watch bridge take injected collaborators instead of reaching for singletons — tracked with per-suite root causes in #190.

PR #211, opened 2026-08-04
mobile: preserve runtime-only thread selection

0xSero:
## Purpose

Replace stale PR #177 with a current-main, parity-safe fix for runtime-only thread launches. Selecting OpenCode or another non-Codex runtime must not fall back to Codex merely because no explicit model override was chosen.

## Key changes

- centralize the iOS runtime/model launch projection and use it in all three launch paths
- preserve the same runtime-only state and request projection on Android
- project the resolved runtime into the initial Rust thread snapshot before the first session-summary projection
- retain the runtime routing map needed for later resume/turn operations
- omit #177’s navigation hunk because that behavior already landed on main

## Verification

- focused Android selection tests pass
- full Android debug unit suite passes
- focused iOS tests pass on the iPhone 17 Pro simulator: 3 tests, 0 failures
- `rustfmt --check` and `git diff --check` pass

## Remaining acceptance

The pristine Codex submodule cannot compile the focused Rust test until the repository patch stack is applied (`JsonRpcWire` drift). CI must exercise the normal sync/patch lane. A physical-device OpenCode creation remains the final end-user acceptance check.

Supersedes #177.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/state/AppLaunchState.kt
+++ b/apps/android/app/src/main/java/com/litter/android/state/AppLaunchState.kt
@@ -33,6 +33,39 @@ data class AppLaunchStateSnapshot(
     val threadPermissionOverrides: Map<String, ThreadPermissionOverride> = emptyMap(),
 )
 
+internal data class AppLaunchModelSelection(
+    val agentRuntimeKind: AgentRuntimeKind?,
+    val model: String?,
+)
+
+internal fun appLaunchModelSelection(
+    modelOverride: String?,
+    selectedModel: String?,
+    selectedAgentRuntimeKind: AgentRuntimeKind?,
+): AppLaunchModelSelection =
+    AppLaunchModelSelection(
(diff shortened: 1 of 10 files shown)


0xSero:
Cleanup review status: the current-main replacement is code-complete and focused iOS/Android tests pass, but it remains draft until the user-facing acceptance surface is proven. Required final check: create a runtime-only OpenCode thread on a live host from both mobile launch paths and confirm the new thread immediately opens and remains projected as OpenCode after refresh/reconnect. The focused Rust test is also awaiting the normal synced Codex patch stack in CI because a bare local submodule hit the pre-existing `JsonRpcWire` incompatibility.

0xSero:
Fable-authorized cleanup: closing for 2.0.2 retriage. Content captured in durable salvage pr-diffs/pr-211.patch (SHA256 f6a591b2...) — head d4d5cc0b, stable patch-id 635f2a8f. No merge. Recoverable from salvage + pull ref. Verified by DeepSeek-V4-Pro.

PR #183, opened 2026-07-20
android: fix realtime voice connection

aj47:
Closes #182.

## Summary

- make Android's initial non-trickle WebRTC offer use one-shot ICE gathering, with a five-second fallback to the best available local description
- make the env API-key fallback construct a clean public OpenAI provider instead of retaining the ChatGPT backend URL
- document the expanded Codex patch and add a regression test for the public endpoint

## Root cause

Android waited indefinitely for ICE gathering to reach `COMPLETE` while using continual gathering. On affected devices, realtime signaling never started. Separately, the ChatGPT-auth API-key fallback reused the ChatGPT backend provider, causing the API key to target a backend call route it cannot access rather than the public Realtime API.

## Verification

- `cd apps/android && ./gradlew :app:testDebugUnitTest`
- `PATH="/Users/ajjoobandi/.rustup/toolchains/1.95.0-aarch64-apple-darwin/bin:$PATH" cargo check -p codex-core --lib`
- reverse patch applicability check against the locally patched Codex submodule
- live Galaxy S25 Ultra (Android 16): SDP answer applied, remote audio track received, ICE reached connected/completed, UI reached listening, speech/transcript updates arrived, and assistant audio played

The public endpoint behavior follows the official [Realtime WebRTC guide](https://developers.openai.com/api/docs/guides/realtime-webrtc).

PR #212, opened 2026-08-04
android: bound realtime voice startup

### LINKED ISSUE #182: Android Realtime voice can hang before connecting
## Problem

Realtime voice on Android can fail before a usable call is established in two ways:

1. The native WebRTC peer uses continual ICE gathering but waits for the gathering state to become `COMPLETE` before sending its non-trickle SDP offer. On affected devices that state never arrives, so `thread/realtime/start` is never called.
2. When a ChatGPT-authenticated Codex session falls back to an environment API key for WebRTC call creation, the fallback clones the current backend provider. That can retain the ChatGPT backend URL instead of creating the call through the public OpenAI Realtime endpoint.

Observed on a Galaxy S25 Ultra running Android 16. The first failure leaves the UI connecting indefinitely; the second surfaces a Realtime call authentication/access error.

## Expected behavior

Android should produce a bounded, complete-enough SDP offer and API-key fallback should create the WebRTC call against `https://api.openai.com/v1`, allowing the public `/v1/realtime/calls` plus sideband flow to connect.

## Proposed fix

- Use one-shot ICE gathering for the initial non-trickle offer and fall back to the current local description after a short timeout.
- Construct a clean public OpenAI provider for API-key WebRTC fallback instead of cloning the ChatGPT-backed provider.
- Add focused regression coverage for the fallback endpoint.

## Verification

Verified end to end on the affected device: SDP answer applied, remote audio track received, ICE reached `CONNECTED`/`COMPLETED`, the UI reached `LISTENING`, speech was transcribed, and spoken model audio played back.

0xSero:
## Purpose

Replace stale PR #183 with a current-main fix for #182: Android realtime voice must not wait indefinitely for ICE completion, and API-key fallback must target the public OpenAI endpoint rather than a ChatGPT backend URL.

## Key changes

- use unified-plan, one-shot ICE gathering and a bounded five-second wait
- continue with the best available local description when ICE completion does not arrive
- expose the peer-connection policy through a focused test seam
- create a clean public OpenAI provider for ChatGPT API-key fallback
- repair all five `RealtimeHandoffRequested` fixtures in the owning server-hint patch

## Verification

- Codex sync/patch stack applies cleanly
- Android compile plus full debug unit suite: 46 tests, 0 failures
- focused Codex provider test: 1 passed, 0 failed
- `git diff --check` passes

## Device evidence

PR #183 documented successful Galaxy hardware verification for the same one-shot/timeout behavior. This current-main replacement has not independently repeated microphone, live ICE, SDP, and authenticated two-way audio on hardware; that remains the final acceptance surface before merge.

Closes #182. Supersedes #183.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/voice/RealtimeWebRtcSession.kt
+++ b/apps/android/app/src/main/java/com/litter/android/voice/RealtimeWebRtcSession.kt
@@ -7,6 +7,7 @@ import android.media.AudioManager
 import android.os.Build
 import android.util.Log
 import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.withTimeoutOrNull
 import org.webrtc.AudioSource
 import org.webrtc.AudioTrack
 import org.webrtc.DataChannel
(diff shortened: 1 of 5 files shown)


0xSero:
Final evidence transfer check: I compared the production `RealtimeWebRtcSession.kt` hunk in hardware-tested #183 with this current-main replacement. The `GATHER_ONCE` policy, five-second bounded wait, timeout fallback, close handling, and local-SDP return path are byte-for-byte behaviorally identical; #212 only extracts the configuration and timeout constant so the policy can be asserted in a unit test. The inherited Galaxy S25 trace therefore exercises the exact production behavior being merged here, while this head separately passes the current Android full suite and shared preparation. I will merge only after the remaining iOS Mobile CI job passes.

PR #214, opened 2026-08-04
assets(android): cut tip images and APK by 2.35 MB

0xSero:
Advances the repository slimming audit with a measured, renderer-validated Android bundle reduction.

## Result

- four Android tip-cat assets: 3,683,479 -> 1,332,872 bytes
- tracked asset reduction: 2,350,607 bytes (63.81%)
- clean current-main debug APK: 118,207,227 bytes
- clean branch debug APK: 115,856,635 bytes
- clean APK reduction: 2,350,592 bytes

The PNGs become lossless WebP under the same Android drawable IDs. No Kotlin or runtime interface changes.

## Fidelity

- dimensions are unchanged for all four assets
- alpha masks are pixel-identical
- every visible RGBA pixel is identical (`visible_pixel_mismatches=0`, `max_visible_delta=0`)
- WebP only normalizes hidden RGB values under fully transparent pixels

## Current-main acceptance

Rebased onto current `main` as commit `15a775a8`.

- clean branch APK SHA-256: `07bc8f833c660060df856952652efa7804e3e6dccaa795d06de2da3a85bd03a8`
- installed branch APK SHA-256: `05f640bb69953f65620eae6da97afadcb0ea0910edfa27dcb74de8e0c04e2420`
- installed and launched on the API 36 `Litter_API_36` emulator
- navigated through Settings -> Tip the Kitty in the installed Compose app
- visually confirmed all four distinct cats render with intact transparency and correct sizing at the $9.99, $24.99, $49.99, and $99.99 thresholds
- `:app:assembleDebug` passed from a clean build
- `:app:testDebugUnitTest` passed
- `git diff --check origin/main...HEAD` passed

## Scope decision

iOS conversion remains excluded because the prior test found that `actool` silently omitted WebP renditions. This PR changes only Android assets and now has its previously missing renderer acceptance.


0xSero:
The remaining iOS job was intentionally canceled after shared-prep and Android both passed. This PR changes Android-only assets, already has a clean local Android assemble/full unit suite, and remains draft for installed Android rendering; the unrelated iOS cross-build was holding the runner needed for #221’s clean-runner hermeticity gate. Rerun the complete matrix before readiness if branch policy requires it; this cancellation is scheduling, not a test failure.

PR #215, opened 2026-08-04
mobile: prefer renamed session titles

### LINKED ISSUE #166: Session list shows preview text instead of renamed title
## Summary

When a session is renamed (e.g., via Claude Code `/rename`), the session list in the sidebar continues to show the first message's raw preview text rather than the user-assigned title.

## Steps to reproduce

1. Connect to a host via kittylitter
2. Start a session — the first message is a command like `/review`, which produces raw XML such as `<command-message>review</command-message>` in the message body
3. Rename the session via Claude Code `/rename` to something meaningful, e.g., `litter-deepseek`
4. Look at the session list in Litter

## Actual behavior

The session list displays the raw first-message preview text:

```
<command-message>review</command-message>...
```

## Expected behavior

The session list should display the user-assigned title (`litter-deepseek`).

## Where the problem is

The title display logic in both platforms prioritizes `preview` over `title`:

**iOS** (`apps/ios/Sources/Litter/Views/SessionsTypes.swift`):
```swift
var displayTitle: String {
    if !preview.isEmpty { return preview }   // preview first
    if !title.isEmpty { return title }
    return "Untitled session"
}
```

**Android** (`apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt`):
```kotlin
get() = preview?.takeIf { it.isNotBlank() }
    ?: title?.takeIf { it.isNotBlank() }    // preview first
    ?: "Untitled session"
```

## Contrast: session detail header does it correctly

The conversation detail header (`AppThreadSnapshot+Runtime.swift`) correctly prioritizes `title` over `preview`:

```swift
var displayTitle: String {
    if !title.isEmpty { return title }       // title first ✅
    if !preview.isEmpty { return preview }
    return "Untitled session"
}
```

And `HomeDashboardSupport.sessionTitle(for:)` also does `title` first.

## Question

Is this a real bug (should title always win over preview in the session list, like the detail header does), or was the `preview > title` ordering intentional for the session list view? The latter seems unlikely given the detail header and dashboard disagree with it, but I wanted to confirm before opening a PR.

0xSero:
Closes #166.

## Root cause

The shared Rust summary already projects an explicit title separately from preview text, but both mobile session-list projections re-selected preview first. A renamed Claude session could therefore keep rendering its first raw command/XML message.

## Change

- prefer trimmed explicit `title` over `preview` in iOS `AppSessionSummary.displayTitle`
- apply the same ordering on Android
- retain preview and `Untitled session` fallbacks
- add explicit title-vs-preview regressions on both platforms

## Verification

- generated current UniFFI bindings from the synced patch stack
- focused Android `SnapshotExtensionsTest`: passed
- Android debug Kotlin compilation: passed
- `git diff --check`: passed

The iOS fast-simulator build/test is still running locally; PR CI is also the final cross-platform gate. This stays draft until those pass.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
+++ b/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
@@ -207,8 +207,8 @@ val AppThreadSnapshot.resolvedPreview: String
     get() = displayTitle
 
 val AppSessionSummary.displayTitle: String
-    get() = preview?.takeIf { it.isNotBlank() }
-        ?: title?.takeIf { it.isNotBlank() }
+    get() = title?.takeIf { it.isNotBlank() }
+        ?: preview?.takeIf { it.isNotBlank() }
         ?: "Untitled session"
 
 val AppThreadSnapshot.contextPercent: Int
--- a/apps/android/app/src/test/java/com/litter/android/state/SnapshotExtensionsTest.kt
+++ b/apps/android/app/src/test/java/com/litter/android/state/SnapshotExtensionsTest.kt
@@ -2,6 +2,9 @@ package com.litter.android.state
 
 import org.junit.Assert.assertEquals
 import org.junit.Test
+import uniffi.codex_mobile_client.AppSessionSummary
+import uniffi.codex_mobile_client.AppSubagentStatus
+import uniffi.codex_mobile_client.ThreadKey
 
 class SnapshotExtensionsTest {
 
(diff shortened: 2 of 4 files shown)


0xSero:
Local cross-platform verification is complete: Android focused unit test/compile passed, and the iPhone 17 Pro simulator ran all 6 `RunningTurnSnapshotTests` with 0 failures, including the new renamed-title regression. Fast simulator Rust/staticlib build and Xcode project generation also passed after installing the pinned Alpine rootfs. Marking ready; GitHub Mobile CI remains the merge gate.

0xSero:
Acceptance complete for this render-only parity fix: Android compiled and passed the focused test; iOS built the exact fast-simulator Rust/app stack and passed all 6 focused tests on the booted iPhone 17 Pro simulator. The change is two ordering swaps with explicit regressions and does not touch shared state. Merging to close #166.

Issue #185, opened 2026-07-20
Fish shell support


maherbeg:
Hi!

I just started setting this up to experiment with. I love how it works at a high level but have had trouble with my setup which is definitely unusual, so feel free to close this.

I'm running
* fish shell
* pi
* on nixos

which means a lot of the traditional pieces for scanning things, or locating binaries may not work. Since fish isn't supported, the path extraction didn't work until I manually added something to `~/.profile` - but it'd be nice to have it work without that.

On this same setup, right now, I get a "Couldn't load this folder" error when adding a new directory for a project. The error details is `Litter.ClientError.Rpc("sever error -32604: failed to spawn command: No such file or directory (os error 2)")`. I'm still hunting through the code to figure out what command is being spawned here and re-creating it in a non-interactive shell.

0xSero:
Confirmed against current main. SSH commands are now forced through a POSIX shell, which avoids parsing fish syntax, but `PROFILE_INIT` still does not import fish or Nix profiles. The folder-load spawn failure also needs to be separated from binary discovery. Please attach the current resolver diagnostics and the specific failing spawned command when available.

0xSero:
Draft PR #216 now covers the proven shell-environment gap: Fish login PATH import plus common Nix profile locations, while generated commands stay under POSIX `sh`. Focused/SSH/full Rust suites pass. A live Fish/NixOS SSH host remains the acceptance gate, and the reported folder-picker `-32604` is explicitly not claimed fixed without host diagnostics.

0xSero:
Shipped in #345 (merged): SSH import now includes fish and Nix profile paths, so login shells configured through them resolve on the shared Rust SSH path. Thanks for the report — closing.

0xSero:
Status update: the shell-environment half of this is merged on main via #345 (`85e5e2df`) — SSH remote setup now imports Fish login PATH plus common Nix profile locations, while generated commands stay POSIX `sh`. Covered by three new PATH-resolution tests in `cargo test -p codex-mobile-client` (790 passing).

Still open from your report: the folder-picker `-32604` error. That one needs host-side diagnostics on a live Fish/NixOS SSH host — if you can share the remote's `$PATH` shape (or the server log line around the picker call), that would unblock reproducing it here. Leaving the issue open for that piece; happy to close if the fish PATH import alone resolves your setup.

PR #216, opened 2026-08-04
ssh: import Fish and Nix profile paths

0xSero:
Addresses the shell-environment portion of #185.

## Root cause

Litter executes generated SSH commands under POSIX `sh`, but its profile importer only tried POSIX/bash/zsh files. A Fish login account—and common NixOS/profile locations—could therefore hide `codex`, Node, and package-manager binaries from the SSH command environment.

## Change

- keep all generated commands under POSIX `sh`
- when Fish is the selected login shell, invoke `fish --login --command` only to export PATH through a sentinel-safe record
- ignore Fish startup chatter and fall back to the POSIX PATH if export fails
- restore common user/system Nix profile bin directories
- preserve existing POSIX/bash/zsh and Node-manager behavior

## Verification

- `/bin/sh -n`
- shellcheck
- focused profile tests: 6/6
- SSH suite: 28 passed, 1 live-only ignored
- full `codex-mobile-client` lib suite: 778 passed, 0 failed, 5 live-only ignored
- `git diff --check`

## Remaining acceptance

No real Fish/NixOS SSH host was available; Fish behavior was exercised through an executable fixture. Keep this draft until a live Fish host proves Codex/package-manager lookup. The reporter's folder-picker `-32604` is a distinct app-server spawn path and is not claimed fixed without host diagnostics.

### DIFF
--- a/shared/rust-bridge/codex-mobile-client/src/ssh/tests.rs
+++ b/shared/rust-bridge/codex-mobile-client/src/ssh/tests.rs
@@ -190,7 +190,8 @@ fn test_profile_init_sources_common_files() {
     assert!(PROFILE_INIT.contains(".zshenv"));
     assert!(PROFILE_INIT.contains(".zprofile"));
     assert!(PROFILE_INIT.contains(".zshrc"));
-    assert!(!PROFILE_INIT.contains("-ic 'printf %s \"$PATH\"'"));
+    assert!(PROFILE_INIT.contains("--login --command"));
+    assert!(PROFILE_INIT.contains("__litter_path_start__"));
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


0xSero:
The remaining Mobile CI run was intentionally canceled after shared-prep and Android passed, to prioritize #221's clean-runner test-hermeticity gate. #216 already has 778 green Rust tests, focused shell/SSH tests, and remains draft for a real Fish/NixOS host. Rerun the platform matrix with that live-host acceptance before readiness; this cancellation is scheduling, not a failure.

0xSero:
Fable-authorized cleanup: closing for 2.0.2 retriage. Content captured in durable salvage pr-diffs/pr-216.patch (SHA256 87031ee2...) — head a8131e26, patch-id b829bc42. No merge. Verified by DeepSeek-V4-Pro.

PR #218, opened 2026-08-04
bridge: restore fork lineage on cold thread lists

### LINKED ISSUE #150: Forked chat/thread becomes inaccessible after leaving session
## Summary

Forking a chat/thread appears to create and open a new fork successfully, but after navigating away from the chat it is not accessible again from KittyLitter/Litter.

## Observed behavior

- Tap the chat/thread **Fork** action.
- Litter creates a new `#` thread/code for the fork and opens it.
- A follow-up message can be sent in the forked chat.
- After leaving the chat/session, the forked chat cannot be found or reopened in KittyLitter/Litter.

## Environments where reproduced

- Local iPhone / "This Device" environment.
- VM environment.

Because this reproduces in both local iPhone and VM env, it seems less likely to be a single local runtime issue and more likely that forked threads are created/opened but not persisted, indexed, rehydrated, or surfaced back into the Sessions list after navigation.

## Expected behavior

A forked chat/thread should remain accessible after leaving the conversation, ideally appearing in Sessions and/or under the Forks filter with enough title/timestamp metadata to distinguish it from the parent thread.

## Notes

The current iOS flow appears to call `forkThread(...)`, set the returned key as active, refresh the new thread snapshot, and open it. That part seems to work initially. The failure appears to happen later when trying to rediscover/reopen the forked thread from Litter after leaving the chat.

Suggested areas to inspect:

- Whether the forked thread is persisted by the local/VM backend after creation.
- Whether thread list/session list APIs include forked threads after `thread/fork`.
- Whether Sessions screen filtering/grouping hides forked threads by default.
- Whether forked threads inherit duplicate titles and are effectively indistinguishable from parent threads.

## User impact

The Fork feature is risky to use because follow-up work in the fork can become inaccessible once the user leaves the chat.

0xSero:
## Purpose

Fix persistent fork rediscovery after leaving a session or restarting Litter. Fixes #150 when merged.

## Root cause

The fork rollout is durable and thread/read returns its forkedFromId, but the upstream local thread-store list projection discarded that immutable field. After a cold restart, thread/list therefore returned the fork without lineage, so the Forks projection could not identify it.

## Changes

- add a narrow Codex patch that restores forked_from_id from the canonical first session metadata record for non-state-DB-only lists
- preserve the state-DB-only contract by performing no filesystem read for those requests
- add a cold-store regression proving a persistent fork is rediscovered with its parent id
- wire the patch into the normal Codex sync and unpatch lanes

## Verification

- cargo test -p codex-thread-store cold_list_threads_preserves_fork_lineage_from_rollout -- --nocapture
- cargo test -p codex-thread-store (73 passed)
- built the patched codex-app-server binary
- end-to-end synthetic acceptance: fork persistent rollout, terminate app-server, start a fresh process, verify thread/list finds the fork with the same forkedFromId, and verify thread/read agrees
- bash -n apps/ios/scripts/sync-codex.sh
- clean apply and reverse-apply checks against pinned Codex 13595c36e218fcbd13df118eeadf00d4eb0e6d31

## Remaining acceptance

A physical iPhone This Device flow and a remote VM flow still need UI-level validation before merge: fork, send a follow-up, leave the conversation, force restart, reopen from Sessions and the Forks filter, and continue the thread. The PR stays draft until those two acceptance surfaces are proven.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,7 +159,8 @@ PATCH_FILES := \
 	$(PATCHES_DIR)/ios-exec-hook.patch \
 	$(PATCHES_DIR)/client-controlled-handoff.patch \
 	$(PATCHES_DIR)/mobile-code-mode-stub.patch \
-	$(PATCHES_DIR)/thread-read-permissions.patch
+	$(PATCHES_DIR)/thread-read-permissions.patch \
+	$(PATCHES_DIR)/thread-list-fork-lineage.patch
 
 BOUNDARY_SOURCES := \
 	$(RUST_DIR)/codex-mobile-client/Cargo.toml \
--- a/apps/ios/scripts/sync-codex.sh
+++ b/apps/ios/scripts/sync-codex.sh
@@ -9,6 +9,7 @@ PATCH_FILES=(
     "$REPO_DIR/patches/codex/ios-exec-hook.patch"
     "$REPO_DIR/patches/codex/mobile-code-mode-stub.patch"
     "$REPO_DIR/patches/codex/thread-read-permissions.patch"
+    "$REPO_DIR/patches/codex/thread-list-fork-lineage.patch"
     "$REPO_DIR/patches/codex/mobile-shell-snapshot-timeout.patch"
     "$REPO_DIR/patches/codex/remote-app-server-websocket-cap.patch"
     "$REPO_DIR/patches/codex/absolute-path-cross-platform.patch"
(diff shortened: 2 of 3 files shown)


0xSero:
The remaining Mobile CI run was intentionally canceled after shared-prep passed, to prioritize #221's clean-runner test-hermeticity gate. #218 already has 73/73 upstream thread-store tests and a real two-process app-server restart probe, and remains draft for physical-iPhone/remote-VM UI acceptance. Rerun the platform matrix with that acceptance before readiness; this cancellation is scheduling, not a failure.

0xSero:
Fable-authorized cleanup: closing for 2.0.2 retriage. Content captured in durable salvage pr-diffs/pr-218.patch (SHA256 e74bb997...) — head ed287e4e, patch-id 8c7455cb. No merge. Verified by DeepSeek-V4-Pro.

Issue #66, opened 2026-03-31
Same remote machine connected via Tailscale + LAN is treated as two servers, duplicating threads and splitting message updates


jelenv:
My prompt to codex (gpt-5.4 high) to investigate this:

> [!NOTE]
>  I noticed there are bugs when you connect to a remote via tailscale and at the same time local LAN. I connected to 2 IP addresses but its one machine. When you are connected like this, it happens that the application litter does not show properly the threads. For example it shows some threads twice or all of them twice and when you send a message in a thread it can happen that the message is not displayed in this thread but it will be displayed in the duplicate one. Please investigate and write me an issue which I can report to the author on GitHub.

What follows is output from codex, edited by me:

### Summary
If the same remote machine is reachable both over Tailscale and over the local LAN, Litter can treat those as two different servers. Once both are connected, thread/session state becomes inconsistent:
- threads can appear twice
- sending a message in one copy of a thread can update the duplicate copy instead of the one currently open

### Steps to reproduce
1. Have one machine reachable from the phone both:
   - by local LAN IP, and
   - by Tailscale IP / MagicDNS
2. Discover/connect to that same machine through both paths so Litter ends up with 2 connections to one physical server.
3. Open the sessions/thread list.
4. Send a message in one of the duplicated threads.

### Actual result
- The same server effectively exists twice.
- Threads can be duplicated.
- Message updates can appear under the duplicate server/thread entry instead of the thread instance that was used to send the message.

### Expected result
The same physical remote machine should collapse to one canonical server identity, even if it is discovered or connected through multiple network paths. Threads should appear once and message updates should always land in the currently selected thread.

### Suspected cause
From code inspection, discovery/session identity is address-based:

- discovered server IDs are created from host/address, e.g. `network-<host>`
- discovery reconciliation dedupes by normalized host only
- thread state is keyed by `(server_id, thread_id)`
- active sessions are reused only by `server_id`

That means one machine on two addresses gets two different `server_id`s, and then the store treats the same upstream threads as separate thread keys.

Relevant code:
- `shared/rust-bridge/codex-mobile-client/src/discovery.rs`
- `shared/rust-bridge/codex-mobile-client/src/types/models.rs`
- `shared/rust-bridge/codex-mobile-client/src/store/reducer.rs`
- `shared/rust-bridge/codex-mobile-client/src/mobile_client_impl.rs`

### Suggested fix direction

[...]

PR #221, opened 2026-08-04
ios: restore full unit test coverage

### LINKED ISSUE #190: iOS test suite is not hermetic: 9 pre-existing failures surfaced by turning on CI tests
#188 turns on `xcodebuild test` for `LitterTests`. That is the first time this suite has ever run in CI — the previous `ios` job stopped at `xcodebuild … build`, which does not even compile the test targets.

Turning it on immediately surfaced 9 failures. **None are caused by #188 or #186** — the code under test (`AppThreadSnapshot+Runtime.swift`, `SavedAppsStore.swift`, `WatchCompanionBridge.swift`) is unchanged since `main`. They are pre-existing and were invisible because nothing ran them.

#188 skips exactly these tests so the rest of the suite can gate. Each needs fixing and unskipping.

## Root cause: the suite is not hermetic

This is the interesting part. The same assertions fail **differently** on a clean runner and on a developer machine, because they read ambient application state rather than fixtures.

`HomeDashboardSupportTests` on GitHub's runner:

```
testHomeDashboardModelRefreshesWhenObservedSnapshotChanges:
  XCTAssertEqual failed: ("[]") is not equal to ("["server-a"]")
```

The same test on my machine:

```
testHomeDashboardModelRefreshesWhenObservedSnapshotChanges:
  XCTAssertEqual failed:
    ("["alleycat:local-studio:3e0bf5506496c8370908487dc6eaf448a6a0a4927a341f56e4f1f35cd9a02312"]")
    is not equal to ("["server-a"]")
```

It returned **my actual paired Local Studio server**. Likewise `testHomeDashboardModelRefreshesRecentSessionsWhenObservedSnapshotThreadChanges` returned a real thread id (`019f9617-b183-7dc2-96c4-9821e2681c84`) instead of the `thread-older` / `thread-newer` fixtures.

So `HomeDashboardModel` is observing the real `AppModel`/`SavedServerStore` singletons instead of the injected snapshot. On a clean machine that reads as empty; on a developer machine it reads as whatever you happen to be paired to. Either way the fixture is ignored.

## The failures

| Test | CI | Local | Cause |
|---|---|---|---|
| `HomeDashboardSupportTests` (7 of 10) | fail | fail | reads real saved-server / thread state instead of the fixture |
| `SavedAppsStoreTests.testPromoteWritesIndexAndHtmlToDisk` | fail | pass | writes to a real container path that does not exist on a clean runner |
| `WatchCompanionBridgeTests.testSnapshotRequestForcesPushThroughTransport` | fail | pass | depends on WatchConnectivity being available/activated |
| `AppSnapshotRuntimeTests.testDisplayModelLabelFallsBackToProviderRuntimeLabel` | fail | fail | expects `"opencode"`, but `displayLabel` falls through to `titlecased` → `"Opencode"` |
| `PerformanceHelpersTests.testTranscriptTurnBuilderCollapsesPreviousTurnOnceANewLiveTurnStarts` | pass | fail | `String(format: "%.1fs")` is locale-sensitive; produced `"3,2s"` |

## Suggested fixes

**`HomeDashboardSupportTests`** — the real work. `HomeDashboardModel` needs its state injected rather than reaching for singletons, so a test can hand it a snapshot and get that snapshot back. Everything else here is small by comparison.

**`SavedAppsStoreTests`** — point the store at a temp directory per test rather than the app container.

**`WatchCompanionBridgeTests`** — inject the transport, or skip when `WCSession.isSupported()` is false rather than asserting a push happened.

**`AppSnapshotRuntimeTests`** — decide which is right. `titlecased` has produced `"Opencode"` since before this test existed, so the assertion looks simply stale; if the intent was that a runtime with no metadata renders its raw id, that is a product change, not a test change.

**`PerformanceHelpersTests`** — worth deciding deliberately: `formatDuration` in `ConversationTimelineView.swift:2492` and `MessageContentBridge.swift:208` both use `String(format: "%.1fs")`, which localizes the decimal separator, while the adjacent `"\(ms)ms"` and `"\(minutes)m \(remainingSeconds)s"` branches do not. So users in comma-decimal locales see `"3,2s"` next to `"450ms"`. Either localize all three or none; then pin the test's locale.

## Why this matters beyond the tests

The suite passing locally for whoever last ran it, while failing on a clean machine, is exactly the failure mode CI exists to catch. Until `HomeDashboardSupportTests` is hermetic, its green result on a developer machine means nothing — it is asserting against that developer's paired servers.

0xSero:
Closes #190

## Purpose
Remove the iOS unit-test quarantine by making the affected tests deterministic and aligning stale assertions with current contracts.

## Key changes
- inject home-dashboard persistence and debounce timing so model tests do not read developer or runner state
- make sub-10-second transcript durations locale-independent
- align saved-app filesystem assertions with the documented direct-root layout
- preserve distinct same-host live server endpoints and the metadata-driven Opencode label contract
- remove all five Mobile CI skip filters

## Verification
- exact Mobile CI-equivalent build-for-testing on iPhone 17 Pro simulator
- all formerly quarantined selectors: pass
- HomeDashboardSupportTests repeated 3x: pass
- PerformanceHelpers locale run with Polish language/region: pass
- full LitterTests target: 211 tests, 0 failures

No submodule changes are included.

### DIFF
--- a/.github/workflows/mobile-ci.yml
+++ b/.github/workflows/mobile-ci.yml
@@ -364,19 +364,6 @@ jobs:
             ONLY_ACTIVE_ARCH=YES \
             build-for-testing
 
-      # The skips below are pre-existing failures uncovered the first time this
-      # suite ever ran in CI. They are not hermetic — they read ambient app
-      # state (saved servers, on-disk stores, WatchConnectivity) or depend on
-      # the runner locale, so they fail differently on a developer machine and
-      # on a clean runner. Tracked in #190; each must be fixed and unskipped
-      # rather than left here. Everything else gates normally.
-      #
-      # HomeDashboardSupportTests and WatchCompanionBridgeTests are skipped as
-      # whole suites rather than test-by-test: their failures come from the
-      # suite's dependency on ambient state, not from individual assertions, so
-      # skipping one test just surfaces the next. Neither reproduces locally —
-      # a developer machine has saved servers and a real WCSession — so the
-      # only way to find the full set would be to keep pushing to CI.
       - name: Test (LitterTests)
         run: |
           xcodebuild \
@@ -387,9 +374,4 @@ jobs:
             ARCHS=arm64 \
             ONLY_ACTIVE_ARCH=YES \
             -only-testing:LitterTests \
-            -skip-testing:LitterTests/HomeDashboardSupportTests \
-            -skip-testing:LitterTests/AppSnapshotRuntimeTests/testDisplayModelLabelFallsBackToProviderRuntimeLabel \
-            -skip-testing:LitterTests/SavedAppsStoreTests/testPromoteWritesIndexAndHtmlToDisk \
-            -skip-testing:LitterTests/WatchCompanionBridgeTests \
-            -skip-testing:LitterTests/PerformanceHelpersTests/testTranscriptTurnBuilderCollapsesPreviousTurnOnceANewLiveTurnStarts \
             test-without-building
--- a/apps/ios/Sources/Litter/Views/HomeDashboardModel.swift
+++ b/apps/ios/Sources/Litter/Views/HomeDashboardModel.swift
@@ -2,6 +2,49 @@ import Foundation
 import Observation
 
 @MainActor
+struct HomeDashboardPersistence {
+    let rememberedServers: () -> [SavedServer]
+    let pinnedKeys: () -> [SavedThreadsStore.PinnedKey]
+    let hiddenKeys: () -> [SavedThreadsStore.PinnedKey]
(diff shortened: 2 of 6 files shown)


0xSero:
Root review complete. The final patch keeps runtime labels metadata-driven, preserves distinct same-host live endpoints per #66, corrects the stale saved-app path assertion (including deletion), and uses injected dashboard persistence rather than ambient developer state. Local acceptance is strong: every former quarantine selector, three HomeDashboard repetitions, Polish-locale duration formatting, and the full 211-test target pass on exact current main. The remaining merge gate is the clean GitHub iOS runner—the surface that originally exposed #190.

PR #223, opened 2026-08-04
ci: make TestFlight cache optional

0xSero:
## Purpose

Keep the standalone TestFlight release runnable when the optional remote sccache credentials are not installed.

## Root cause

Run 30951375473 passed the live App Store Connect preflight, then failed before compilation because `SCCACHE_R2_ENDPOINT` and the R2 access secrets were empty while the workflow still forced `RUSTC_WRAPPER=sccache` and unconditionally started the remote cache.

## Changes

- configure sccache only when the endpoint and both credentials are present
- skip setup/priming when the cache is unavailable
- let Cargo run directly instead of forcing an unusable wrapper

This matches the existing optional-cache behavior in the merged Mobile Release lane.

## Verification

- parsed the workflow as YAML
- exercised both cache-enabled and cache-disabled configuration paths from the workflow source
- Actionlint passed, ignoring only the pre-existing SC2129 style finding elsewhere in the workflow
- `git diff --check`

### DIFF
--- a/.github/workflows/ios-testflight.yml
+++ b/.github/workflows/ios-testflight.yml
@@ -87,10 +87,26 @@ jobs:
         with:
           targets: aarch64-apple-ios,aarch64-apple-ios-sim,x86_64-apple-ios,i686-unknown-linux-musl
 
+      - name: Configure sccache
+        run: |
+          if [ -n "${SCCACHE_ENDPOINT:-}" ] && [ -n "${AWS_ACCESS_KEY_ID:-}" ] && [ -n "${AWS_SECRET_ACCESS_KEY:-}" ]; then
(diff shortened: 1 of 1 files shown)


0xSero (commented):
Self-review complete: no blocking findings.

I traced both environment branches from the workflow source:

- with endpoint + access key + secret key, the workflow sets `RUSTC_WRAPPER=sccache` and runs setup/priming;
- if any value is absent, it clears the remote-cache environment, skips setup/priming, and leaves Cargo without a forced wrapper.

The change does not log secret values, does not affect signing or App Store credentials, and is limited to the standalone TestFlight preparation job. YAML parsing, both executable configuration-path assertions, Actionlint (apart from the pre-existing SC2129 style finding), and `git diff --check` pass.

PR #224, opened 2026-08-05
ci: normalize Apple signing identity imports

0xSero:
## Purpose

Make Apple release preflight exercise the same macOS Keychain surface used by archive jobs, and make the standalone TestFlight lane complete its manual signing inputs.

## Root cause

Run 30951914596 passed OpenSSL P12 validation and completed the 82-minute native release build, then macOS Keychain rejected the P12 with `SecKeychainItemImport: MAC verification failed during PKCS12 import`.

OpenSSL can verify and extract the identity, while Keychain can import the extracted PEM. The failure is therefore a PKCS#12 import-compatibility boundary, not a bad certificate or missing private key.

The same review found that the standalone lane omitted the Watch and Watch Complications provisioning profiles required by its manual signing mode.

## Changes

- add one shared Apple signing import helper that verifies/extracts the P12 with OpenSSL and imports the PEM into Keychain
- use it in standalone TestFlight and Mobile Release
- make Apple preflight create a disposable Keychain, import the identity, and require a valid codesigning identity before native builds begin
- validate and install all four iOS/watchOS App Store profiles in the standalone lane

## Verification

- generated a temporary certificate/private key/P12, imported it through the shared helper, and resolved the certificate from a disposable macOS Keychain
- verified the helper rejects missing password inputs
- `bash -n` and ShellCheck pass
- both workflows parse as YAML
- Actionlint passes with only the existing SC2129 style exclusion
- `git diff --check`

### DIFF
--- a/.github/workflows/ios-testflight.yml
+++ b/.github/workflows/ios-testflight.yml
@@ -45,6 +45,9 @@ jobs:
           IOS_DIST_CERT_PASSWORD: ${{ secrets.IOS_DIST_CERT_PASSWORD }}
           IOS_APP_STORE_PROFILE_B64: ${{ secrets.IOS_APP_STORE_PROFILE_B64 }}
           IOS_LIVE_ACTIVITY_APP_STORE_PROFILE_B64: ${{ secrets.IOS_LIVE_ACTIVITY_APP_STORE_PROFILE_B64 }}
+          IOS_WATCH_APP_STORE_PROFILE_B64: ${{ secrets.IOS_WATCH_APP_STORE_PROFILE_B64 }}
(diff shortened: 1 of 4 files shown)


0xSero (commented):
Self-review complete: no blocking findings.

Review covered the exact failure boundary and adjacent manual-signing inputs:

- the P12 password stays in an environment variable and is never placed on an OpenSSL command line;
- the extracted private key is created with owner-only permissions and removed by an EXIT trap;
- the disposable preflight Keychain must expose at least one valid codesigning identity, so an importable-but-incomplete certificate cannot pass;
- standalone TestFlight now validates and installs the app, live activity, watch app, and watch complications profiles required by manual signing;
- Mobile Release and standalone TestFlight use the same import helper, eliminating drift.

Local macOS integration testing imported a generated key/certificate through the helper into a disposable Keychain, resolved the certificate, and verified missing password inputs fail. Bash syntax, ShellCheck, YAML parsing, Actionlint, and diff checks pass.

PR #235, opened 2026-08-05
fix: preserve Local Studio session identity

0xSero:
## Purpose

Keep Local Studio sessions visible and correctly labeled across mobile reconnects and scanner refreshes, without changing ordinary Codex or generic multi-agent behavior, and prepare the validated 2.0.0 mobile release.

## Changes

- route Local Studio-only session listing and new thread starts explicitly to the Local Studio runtime
- reconcile Local Studio list pages additively so partial pages cannot prune durable sessions
- preserve Local Studio identity when notifications arrive before asynchronous thread insertion
- keep current Local Studio sessions visible alongside legacy pins on both mobile platforms
- skip OpenAI account probes and sign-in warnings for Local Studio-only connections
- force Pi and Local Studio Pi to never request approvals and to use full filesystem access for start, resume, fork, and turns
- preserve existing Codex permission and generic scanner behavior
- set iOS and Android release version to 2.0.0

## Verification

- shared Rust: 797 total, 792 passed, 5 environment-dependent tests ignored, 0 failed
- Android unit tests: 52 passed, 0 failed
- iOS unit tests: 216 passed, 0 failed
- iOS UI tests: 7 executed, 0 failed, 1 intentional live-screenshot skip
- real Kittylitter/Iroh Local Studio E2E: 20 passed, 0 failed
  - exact streamed response and reasoning/item lifecycle
  - tool calling, shell execution, file write/read ordering, and full filesystem access
  - reconnect cursor and mid-turn disconnect hydration
  - desktop-to-phone resume and phone-to-desktop visibility
  - compaction and post-compaction continuation
- Android emulator upgrade from persisted 1.6 data to 2.0.0
  - existing sessions remained visible and Local Studio-labeled
  - new thread routed through Local Studio Pi
  - streamed exact response remained after background/foreground and reinstall
  - no Local Studio account/read probe after reconnect
- exact native versions verified: 2.0.0 / build 200000254

## Companion fix

- Local Studio packaged Pi provider-data fix: sybil-solutions/local-studio#358

### DIFF
--- a/apps/android/app/build.gradle.kts
+++ b/apps/android/app/build.gradle.kts
@@ -34,7 +34,7 @@ android {
         minSdk = 26
         targetSdk = 36
         versionCode = 200000254
-        versionName = "1.7.0"
+        versionName = "2.0.0"
         buildConfigField("boolean", "ENABLE_ON_DEVICE_BRIDGE", "true")
         buildConfigField("String", "RUNTIME_STARTUP_MODE", "\"hybrid\"")
         buildConfigField("String", "APP_RUNTIME_TRANSPORT", "\"app_bridge_rpc_transport\"")
--- a/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
+++ b/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
@@ -96,7 +96,8 @@ val AppServerSnapshot.connectionProgressDetail: String?
 val AppServerSnapshot.statusLabel: String
     get() = when {
         connectionProgressLabel != null -> connectionProgressLabel!!
-        transportState == AppServerTransportState.CONNECTED && !isLocal && account == null -> "Sign in required"
+        transportState == AppServerTransportState.CONNECTED && requiresOpenaiAuth && account == null ->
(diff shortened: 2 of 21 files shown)


0xSero (commented):
Self-review complete on the exact head (`6e8aae7c`). I reviewed the shared Rust routing/reconciliation/permission changes and both native projections. I found one scope leak in the first pass: the reconnect account-probe skip applied to every non-Codex-only runtime. That is fixed in `6e8aae7c`; the skip is now limited to a sole `local-studio` runtime, while Codex, Pi, mixed, and legacy inventories retain the existing probe behavior.

No remaining actionable findings. Exact-head verification: Rust 797 total / 792 passed / 5 environment-dependent ignored / 0 failed; Android 52 passed / 0 failed; iOS 216 unit + 7 UI / 1 intentional live-screenshot skip / 0 failed; `git diff --check` clean; fresh installed Android 2.0 Local Studio session streamed the exact answer with thread/start, turn/start, list, resume, and hydration all routed to Local Studio; background/foreground and upgrade persistence passed; Local Studio end-to-end proof 20/20 passed.

PR #278, opened 2026-08-12
docs: remove duplicate agent guide

0xSero:
## Purpose
Remove a byte-for-byte duplicate contributor/agent instruction file so there is one source of truth.

## Changes
- delete `CLAUDE.md`, which exactly duplicated `AGENTS.md`
- update its sole source-comment reference to `AGENTS.md`

## Verification
- confirmed `AGENTS.md` and `CLAUDE.md` had identical SHA-256 content before removal
- `cargo check -p codex-mobile-client`

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1 +1,0 @@
-AGENTS.md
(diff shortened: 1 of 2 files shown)


0xSero:
Withdrawn: `CLAUDE.md` is an intentional symlink to `AGENTS.md`, not a second stored copy. Keeping the compatibility alias is the correct cleanup decision.

Issue #283, opened 2026-08-13
bug: Never receive response from 1st message in a new thread when using OpenCode


DatScreamer:
Reproduction:
1. Connect to server
2. Click + button to create a new thread
3. select model / project if needed
4. send literally any message
5. thread is created, "New session [timestamp]"
6. You can wait as long as you want and nothing will ever return from that initial message.

- The "Loading conversation" shows up and then goes away but nothing appears, at all. It's just an empty page.
- Send another message, and actually get your first response in that thread.When opening the thread before sending a 2nd message doesn't even show the 1st message.

Also when cancelling any turn, I get this error:
Litter.ClientError.Rpc("server error -32600: thread not found: [uuid here]")

Evidence (2nd turn in new convo):
<img width="381" height="185" alt="Image" src="https://github.com/user-attachments/assets/0becd2b2-5fea-4ff5-8b72-e18f45d41c00" />

AI provider: OpenCode
AI plan: OpenCode Go (free models also don't work)
OpenCode version: headless OpenCode running v1.18.16

My headless ChatGPT via Litter works perfectly, and is self-hosted on the exact same headless container.

DatScreamer:
Pls sir throw some of your glorious TFLOPs at this

DatScreamer:
Opened a fix PR for this: https://github.com/0xSero/litter/pull/284

Root cause was two client-side defects in the Rust client, both specific to multi-runtime hosts (my harness serves both an OpenCode runtime and the ChatGPT/codex app-server):

1. turn/interrupt routed to the wrong runtime — runtime_for_request did not match TurnInterrupt, so cancel fell through to the default codex channel and hit the ChatGPT app-server, which does not know the OpenCode thread id -> server error -32600: thread not found. ChatGPT-via-Litter only worked because on a codex-only connection the default channel is the right one.

2. Brand-new thread evicted by a thread/list sync — sync_thread_list_for_runtime / finalize_thread_list_sync evict any thread absent from the incoming page (unless active/local-studio). A just-started OpenCode session is not returned by GET /session until its first message lands, so the snapshot was evicted mid-first-message and mutate_thread_with_result / overlay staging silently dropped the optimistic bubble and every streamed event. A second message re-materializes the thread via history hydration, which is why sending a second message returned both responses.

PR #284 routes TurnInterrupt by owning runtime and keeps recently-created/in-flight threads across list syncs, with regression tests.

PR #284, opened 2026-08-13
fix(opencode): route turn/interrupt to owning runtime; keep new threads across list sync

DatScreamer:
Fixes the first-message-in-a-new-thread failure when using the **OpenCode** provider (0xSero/litter#283). The ChatGPT/codex provider is unaffected.

## Root cause

Two client-side defects, both specific to multi-runtime hosts (a host that serves both an OpenCode runtime and a ChatGPT/codex app-server, e.g. the headless `agent-harness`):

**1. `turn/interrupt` routed to the wrong runtime** — `runtime_for_request` (mobile_client/event_loop.rs) matched every thread-scoped request except `TurnInterrupt`, so it fell through to the default `"codex"` channel. Cancelling a turn therefore landed on the ChatGPT app-server, which doesn't know the OpenCode thread id and replied `server error -32600: thread not found`. ChatGPT-via-Litter "works" only because on a codex-only connection the default channel happens to be the right one.

**2. Brand-new thread evicted by a `thread/list` sync** — `sync_thread_list_for_runtime` / `finalize_thread_list_sync` evict every thread not in the incoming page unless it's the active thread or local-studio. A just-started OpenCode session is not returned by `GET /session` until its first message lands, so during that window the snapshot is evicted. With the snapshot gone, `mutate_thread_with_result` and `stage_local_user_message_overlay` silently no-op — the optimistic user bubble AND every streamed turn event are dropped. That's why the first message shows nothing at all, and why a second message (which re-materializes the thread via history hydration) "returns both".

## Fix

- `runtime_for_request`: route `TurnInterrupt` by the thread's owning runtime (same as `TurnStart`/`TurnSteer`); map its wire method in `client_request_wire_method`.
- reducer: never evict a thread that has an in-flight turn, queued follow-ups, a staged local overlay, or was created within the last minute — closing the brand-new-thread eviction window.
- Tests: `turn_interrupt_routes_to_owning_runtime_not_codex`, `client_request_wire_method_maps_turn_interrupt`, `sync_thread_list_preserves_recently_created_thread_not_in_incoming_page`, and a control test `sync_thread_list_evicts_old_inactive_thread_not_in_incoming_page`.

Verified server-side that opencode serve handles a fresh session + first prompt correctly (2s response via direct HTTP repro), confirming the bug lives in the client, not the harness/server.

### DIFF
--- a/shared/rust-bridge/codex-mobile-client/src/mobile_client/event_loop.rs
+++ b/shared/rust-bridge/codex-mobile-client/src/mobile_client/event_loop.rs
@@ -356,7 +356,7 @@ impl MobileClient {
         })
     }
 
-    fn runtime_for_request(
+    pub(crate) fn runtime_for_request(
         &self,
         server_id: &str,
         request: &upstream::ClientRequest,
@@ -386,6 +386,12 @@ impl MobileClient {
             }
             upstream::ClientRequest::TurnStart { params, .. } => Some(params.thread_id.as_str()),
             upstream::ClientRequest::TurnSteer { params, .. } => Some(params.thread_id.as_str()),
+            // turn/interrupt must go to the runtime that owns the thread.
+            // Without this arm it falls through to the `codex` default channel
+            // and, on a multi-runtime host, the cancel lands on the ChatGPT
+            // app-server which doesn't know the OpenCode thread id
+            // (`server error -32600: thread not found`).
+            upstream::ClientRequest::TurnInterrupt { params, .. } => Some(params.thread_id.as_str()),
             _ => None,
         };
         thread_id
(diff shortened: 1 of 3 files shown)


0xSero:
Closing: the correctness fix from this PR (by @DatScreamer) has landed on `main` via PR #289 (rebase-merged as commit a8fb7bfb2e0c5b17ad72757c7036bea82c61331e — https://github.com/0xSero/litter/commit/a8fb7bfb2e0c5b17ad72757c7036bea82c61331e). The only deviation from the original PR head was a two-line test-only adaptation adding `request_id: upstream::RequestId::Integer(crate::next_request_id())` to the two `ClientRequest::TurnInterrupt` test constructions, required because current main's upstream protocol mandates `request_id` (the original fork-CI E0063). Production code is unchanged. Post-merge Mobile CI run 31739779029 passed all 3 jobs (shared-prep, ios, android) on a8fb7bfb. Branch/worktree retained.

PR #298, opened 2026-08-14
mobile: support current Amp modes

AmAzing129:
## Purpose

Update Litter's Amp model picker and fallback catalog for Amp's current `low`, `medium`, `high`, and `ultra` modes, replacing the retired `smart`, `rush`, and `deep` modes.

Depends on https://github.com/0xSero/alleycat/pull/47. Once that PR merges, this branch should refresh the pinned Alleycat revision before becoming ready for review.

## Changes

- expose the four current Amp modes and default to `medium`
- stop presenting a separate reasoning-effort selector for Amp modes
- filter Android's model picker against the current mode names
- update Amp thread fixtures and mode normalization tests

## Verification

- `mise x rust@1.95.0 -- cargo test --manifest-path shared/rust-bridge/Cargo.toml -p codex-mobile-client amp_` (6 passed)
- `git diff --check`

Android unit tests were not run locally because an Android Studio JBR/JDK is not installed in this environment.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/ui/common/ModelSelectorPanel.kt
+++ b/apps/android/app/src/main/java/com/litter/android/ui/common/ModelSelectorPanel.kt
@@ -515,7 +515,7 @@ internal fun effortLabel(value: ReasoningEffort): String = when (value) {
 private fun ModelInfo.defaultReasoningEffortSelection(): String? =
     if (supportedReasoningEfforts.isEmpty()) null else effortLabel(defaultReasoningEffort)
 
-private val AmpVisibleModes = setOf("smart", "rush", "deep")
+private val AmpVisibleModes = setOf("low", "medium", "high", "ultra")
 
 private fun normalizedAmpModeName(value: String): String =
     value.trim()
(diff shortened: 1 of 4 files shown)


0xSero:
Blocked on a pin refresh: alleycat#47 is merged now, so this can repin to 0xSero/alleycat main and drop the fork reference. Then it is mergeable.

AmAzing129:
Thanks! AmAzing129 checked the dependency pin before updating it. Litter currently pins Alleycat to `417f2a9fe38cbed63754f0af7df61f32ec3034e6`, which is the tip of `release/litter-mobile-v1.6-base`, rather than `main`.

PR #47 is merged into `main`, but the production branch and `main` have diverged. Comparing the pinned tree directly with `main`, AmAzing129 found that `main` lacks functionality present in the production version, including the Local Studio implementation, the `local-studio` / `studio` CLI commands, and Pi file-search support. So repinning to `main` would be a broader dependency migration, not just picking up the Amp mode fix.

Could you handle or confirm the intended migration from the production branch to `main` separately? AmAzing129 would prefer to keep this PR scoped to supporting Amp’s `low` / `medium` / `high` / `ultra` modes, without making decisions about those other features. Once the dependency baseline is settled, AmAzing129 can update this PR accordingly.

0xSero:
Superseded by the rebranch PR (see the new PR from rebranch/amp-current-modes): rebased onto current main, review findings applied (catalog entry modes/effort-lock, seeding test, retired-mode replacement test, doc comment), and the redundant iOS filter-removal commit dropped since main's filter is already catalog-driven. Thanks for the contribution — the Rust plumbing was correct as reviewed.

PR #302, opened 2026-08-15
bridge: attach to a running opencode server instead of respawning

### LINKED ISSUE #301: opencode SSH bridge respawns a new server on every reconnect instead of attaching to the running one
## Summary

On every reconnect to an OpenCode SSH server, the app spawned a **fresh `opencode serve` on a random port** and killed the old one on close. The freshly-spawned server had to re-materialize its threads from disk, so the Sessions list appeared empty for a while, and old servers accumulated if a previous one wasn't torn down cleanly.

## Fix (PR)

`fd010cec` mirrors the Codex bootstrap pattern:
- Probe opencode's well-known port range (starting at 4096) for an existing healthy server (`/global/health`) and **attach** to it instead of launching.
- Only spawn a fresh server on a free candidate when nothing is running.
- The spawned server persists across disconnects, and the session dir is stable (`opencode/`, not `opencode-<millis>`) so the pidfile doubles as the ownership marker.
- Kill-on-close cleanup is dropped in favor of the existing failure-path cleanup.

## Verification (manual, iOS simulator against the agent-harness guest)

- First connect spawns exactly one `opencode serve` on 4096 with `~/.litter/sessions/opencode/{out.log,err.log,agent.pid}`.
- Reconnect **attaches** to the existing process (`ssh bridge opencode attach existing port=4096`) — same PID, no respawn.
- Exactly one `opencode serve` process on port 4096 throughout; no accumulating `opencode-<millis>` session dirs.

## Follow-up found during testing

The app's opencode bridge only tracks sessions it created itself — it never calls `GET /session` to enumerate the server's pre-existing sessions, so externally-created opencode sessions don't appear in the app's Sessions list. That's tracked separately.

DatScreamer:
Closes #301

## Summary
Reconnect used to spawn a fresh `opencode serve` on a random port every time and kill the old one on close, which left the session list empty while the new server re-materialized its threads. Mirror the Codex bootstrap instead: probe opencode's well-known port (4096) range, attach to an existing healthy server, and only spawn on a free candidate when nothing is running.

## Changes
- Probe the well-known port range (4096+) for an existing healthy `opencode serve` (`/global/health`) and attach to it instead of launching.
- Spawn a fresh server only on a free candidate; the spawned server persists across disconnects.
- Stable session dir (`opencode/`, not `opencode-<millis>`) so the pidfile doubles as the ownership marker.
- Kill-on-close cleanup dropped in favor of the existing failure-path cleanup.

## Verification (manual, iOS simulator against the agent-harness guest)
- First connect spawns exactly one `opencode serve` on 4096 with `~/.litter/sessions/opencode/{out.log,err.log,agent.pid}`.
- Reconnect attaches to the existing process (`ssh bridge opencode attach existing port=4096`) — same PID, no respawn.
- Exactly one `opencode serve` process on port 4096 throughout; no accumulating `opencode-<millis>` session dirs.

## Follow-up (tracked separately)
The app's opencode bridge only tracks sessions it created; it never calls `GET /session` to enumerate pre-existing opencode sessions, so externally-created sessions don't appear in the app's Sessions list.


0xSero:
Flagging scheduling: #319 has merged (`0824605d`) and it touches `ssh_bridge.rs` — the agent catalog replaced four scattered hardcoded runtime allowlists with a single source, changing `probe_remote_agents` and the reconnect ordering. This PR's attach-instead-of-respawn approach is still right, but it will likely need a rebase now.

PR #304, opened 2026-08-15
fix(status): don't gate connected servers on opencode OpenAI sign-in

### LINKED ISSUE #303: Connected opencode SSH server stays yellow "Sign in required" due to codex runtime auth gate
## Summary

A connected ssh-bridge server that exposes both `opencode` and `codex` runtimes shows a **yellow "Sign in required"** status dot even though the transport is connected and opencode is fully usable.

## Root cause

The status dot is gated on:

```
transportState == .connected && requiresOpenaiAuth && account == nil  → pending (yellow)
```

`requiresOpenaiAuth` / `account` come from the server's `account/read`, which the app routes to the **codex** runtime. When codex on the remote host isn't logged into OpenAI, it reports `requiresOpenaiAuth=true, account=nil` — so the *whole server* is gated on codex's missing login, even when an **opencode** runtime is available and manages its own server-side credentials (its configured provider).

Confirmed via a temporary diagnostic log while connected:

```
[YELLOWDBG] server=ssh-bridge:agent-harness.lan transport=connected step=nil label=nil needsOpenaiAuth=1 hasAccount=0
```

(transport connected, no stuck progress; the only yellow trigger was the auth gate.)

## Fix

Add a `needsAppOpenaiSignIn` projection that exempts servers with an **available opencode runtime**:

```
needsAppOpenaiSignIn = connected && requiresOpenaiAuth && account == nil
                        && !(agentRuntimes contains kind == "opencode" && available)
```

Applied to `statusLabel`, `statusColor`, and `statusDotState` on both iOS (`AppServerSnapshot+UI.swift`) and Android (`SnapshotExtensions.kt`). A connected server with opencode available renders green; codex-only servers still prompt "Sign in required".

## Verification
- Manual: connected to an opencode SSH server on iOS simulator (and Android emulator) — status now green while connected.

DatScreamer:
Closes #303

## Summary
A connected ssh-bridge server exposing both `opencode` and `codex` runtimes stayed yellow ("Sign in required") because the status dot is gated on `requiresOpenaiAuth && account == nil` — values sourced from the **codex** runtime's `account/read`. When codex on the remote host isn't logged into OpenAI, it forces the whole server yellow even though an **opencode** runtime is available and manages its own server-side credentials.

## Changes
- New `needsAppOpenaiSignIn` projection: `connected && requiresOpenaiAuth && account == nil && !(agentRuntimes contains kind=="opencode" && available)`.
- Applied to `statusLabel`, `statusColor`, `statusDotState` on iOS (`AppServerSnapshot+UI.swift`).
- Same projection applied on Android (`SnapshotExtensions.kt`).

## Behavior
- Connected server with an available opencode runtime → **green** (opencode works without the app's OpenAI account).
- Codex-only server (or one genuinely needing the app's OpenAI login) → still "Sign in required" (yellow).

## Verification
- Manually tested on iOS simulator and Android emulator against an opencode SSH server: status renders green while connected.

### DIFF
--- a/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
+++ b/apps/android/app/src/main/java/com/litter/android/state/SnapshotExtensions.kt
@@ -90,10 +90,23 @@ val AppServerSnapshot.connectionProgressLabel: String?
         null -> null
     }
 
+/**
+ * True when the server reports it needs the app's OpenAI account to be useful
+ * and no account is configured. Servers with an available opencode runtime are
+ * exempt: opencode holds its own credentials on the server (its configured
+ * provider), so the app's OpenAI sign-in isn't required to use it.
+ */
+val AppServerSnapshot.needsAppOpenaiSignIn: Boolean
+    get() = if (transportState != AppServerTransportState.CONNECTED || !requiresOpenaiAuth || account != null) {
+        false
+    } else {
+        !agentRuntimes.any { it.kind == "opencode" && it.available }
+    }
+
 val AppServerSnapshot.statusLabel: String
     get() = when {
         connectionProgressLabel != null -> connectionProgressLabel!!
-        transportState == AppServerTransportState.CONNECTED && requiresOpenaiAuth && account == null ->
+        needsAppOpenaiSignIn ->
             "Sign in required"
         else -> transportState.displayLabel
     }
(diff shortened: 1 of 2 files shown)


DatScreamer:
I was actually wrong about this.

Issue #306, opened 2026-08-15
bug: Large conversations render unbounded (crash / blank surface) on iOS; older history loads via a manual button


DatScreamer:
Reproduction:
1. Open a conversation with many turns (e.g. a large OpenCode or codex thread).
2. On iOS, the app can block the main thread until the watchdog kills it (0x8BADF00D), or render a blank conversation surface.
3. Scrolling up to read older messages requires tapping "Load earlier messages" repeatedly, with small (5-turn) pages.

Root cause:
The client issues metadata-only thread/read requests (include_turns=false) in the shared multi-runtime resume/fallback path. Some bridges — including the opencode bridge — return the full turn archive even for metadata-only reads. The client merged all embedded turns unconditionally, so a 5-turn page could silently become hundreds of UI items rendered at once, exceeding SwiftUI's layout budget and blocking the main thread.

Additionally, iOS had no scroll integration for history loading: a manual button, page size 5, and a "Latest" jump that could be swallowed by a decelerating ScrollView.

Fix (two PRs):
- #PR1: honor include_turns=false so metadata reads stay bounded (shared Rust, all runtimes), plus pin the alleycat fork's pagination commits.
- #PR2: replace the manual button with scroll-triggered infinite scroll, larger pages, a working "Latest" jump, and auto-collapse for large conversations (iOS).

DatScreamer:
Fix split across two PRs:

- **#307** `fix(thread): keep metadata thread reads paginated` — shared Rust, runtime-agnostic: honors `include_turns=false` so metadata reads stay bounded; pins the alleycat fork's pagination commits.
- **#308** `feat(ios): paginate older turns with infinite scroll` — iOS: scroll-triggered older-page prefetch, page size 5→20, working "Latest" jump, auto-collapse past 200 items.

DatScreamer:
Server-side dependency now upstreamed: 0xSero/alleycat#49 (closes 0xSero/alleycat#48) — pins the opencode bridge pagination + session enumeration that #307 consumes.

0xSero:
Partially fixed in #319 (merged as `0824605d`).

**Fixed:** the unbounded-render root cause. `c153d1e5` makes `include_turns=false` authoritative, so a compatibility bridge returning a full thread archive for a metadata-only read can no longer replace a bounded page with hundreds of UI items. Turn page size also went 5 → 20, which removes the "tap Load earlier 40 times" problem — safe only because of the `include_turns` fix above.

[...]

PR #307, opened 2026-08-15
fix(thread): keep metadata thread reads paginated

DatScreamer:
Closes part of #306

## Problem
Metadata-only `thread/read` (include_turns=false) is issued in the shared multi-runtime resume/fallback path. Some bridges (incl. the opencode bridge) return the full turn archive regardless. The old reconcile handler (`apply_thread_read_response`) merged all embedded turns unconditionally, so a 5-turn page could be replaced by hundreds of UI items rendered at once — blocking the main thread until the watchdog killed the app (0x8BADF00D).

## Changes
- `thread/read` reconciliation now reads the request's `include_turns` flag and clears embedded turns for metadata-only reads, so `apply_pagination_merge` preserves the existing paged items + cursor.
- Pin alleycat bridges to `55f78eca` (aa391b5 + 6 commits: native-message cursor pagination, idle/active status, keep metadata reads paginated, project-aware session enumeration).
- `update-alleycat-main.sh`: treat any rev-pinned git dep as pinned so the fork pin survives the CI shared-prep step (previously the script only recognized `dnakov/alleycat`, so CI's `cargo update --precise` overrode the fork pin with upstream main and broke the build).

## Scope
Shared Rust, runtime-agnostic: benefits codex, claude, pi, local-studio, and opencode. The fork pin bumps all four bridge crates; only the opencode bridge's behavior differs.

## Tests
- New reconcile test: `metadata_only_thread_read_ignores_bridge_embedded_history`.

## Verification
- `make rust-test`
- iOS emulator


DatScreamer:
Server-side dependency: 0xSero/alleycat#49 (issue 0xSero/alleycat#48). The Cargo pin in this PR points at fork rev `55f78eca`, which is the head of that PR.

0xSero:
The `reconcile.rs` half of this landed in #319 (merged as `0824605d`) as `c153d1e5` — `include_turns=false` is now authoritative, so a metadata-only read cannot smuggle in a full turn archive. That change is runtime-agnostic and stands alone.

Holding the dependency half. It repoints all five `alleycat-*` crates from `dnakov/alleycat@417f2a9f` to `DatScreamer/alleycat@eead5e5a`. I checked whether the URL could just be rewritten to the canonical remote — it can't: `dnakov/alleycat` and `0xSero/alleycat` are identical mirrors (same branches, same SHAs) and `eead5e5a` is the head of neither; it resolves only through fork-network object sharing.

It also relaxes the pin guard in `tools/scripts/update-alleycat-main.sh` from `dnakov/alleycat\.git.*rev = ` to `\.git.*rev = `, which would let any future fork URL pass the "is pinned" check.

Keeping this open for the canonical revision to land upstream.

0xSero:
The reconcile change is good, but this repins alleycat to a fork. alleycat#49 needs to land upstream first, then repin to 0xSero/alleycat.

PR #308, opened 2026-08-15
feat(ios): paginate older turns with infinite scroll

DatScreamer:
Closes part of #306
Depends on #307 (land after it so the paged cursor behaves predictably).

## Problem
iOS conversation history loading was manual (a "Load earlier messages" button), page size 5, and the "Latest" jump button could be consumed just by stopping a decelerating ScrollView.

## Changes (ConversationView, AppModel, TranscriptTurn, Settings, tests)
- Scroll-triggered prefetch: tracks visible turn IDs and loads the next older page as the user approaches the top, debounced per cursor so concurrent prefetches collapse.
- Page size 5 → 20; `loadOlderTurns` now returns whether a page loaded.
- "Latest" re-routed through semantic `ScrollPosition` + a UIKit scroll-view resolver so the jump executes even while momentum is decelerating.
- Merged exploration rows anchored to their newest constituent turn so prepending older pages keeps scroll identity stable.
- Auto-collapse conversations past 200 items.

## Scope
iOS only; the conversation surface is shared across all runtimes, so this improves every provider (codex, claude, pi, local-studio, opencode).

## Parity note
Android still uses a manual "Load earlier messages" button (ConversationScreen.kt). The Rust metadata-read fix (#307) benefits both; the infinite-scroll UX is iOS-only in this change. Android parity is a follow-up.

## Tests
- 3 new PerformanceHelpersTests: auto-collapse, prefetch-before-cache-edge, exploration identity.

## Verification
- `make ios-sim-fast`

### DIFF
--- a/apps/ios/Litter.xcodeproj/project.pbxproj
+++ b/apps/ios/Litter.xcodeproj/project.pbxproj
@@ -1966,12 +1966,12 @@
 			projectRoot = "";
 			targets = (
 				A3D84F588EDBC1EA550D6691 /* Litter */,
-				4D58398238C075D3A3A3AB9B /* LitterLiveActivity */,
 				6EBBC661C017816B2D4245DF /* LitterMac */,
-				DA07A1EAC6173D09F6FB99D9 /* LitterTests */,
-				7BB0106D190F10717D781234 /* LitterUITests */,
+				4D58398238C075D3A3A3AB9B /* LitterLiveActivity */,
 				3B991C03E72B1546CAD35698 /* LitterWatch */,
 				C0C98C6012493B73CDD5AD6E /* LitterWatchComplications */,
+				DA07A1EAC6173D09F6FB99D9 /* LitterTests */,
+				7BB0106D190F10717D781234 /* LitterUITests */,
 			);
 		};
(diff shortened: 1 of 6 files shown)


DatScreamer:
Depends on the server-side pagination from 0xSero/alleycat#49 (issue 0xSero/alleycat#48), pinned via #307's Cargo.lock bump.

0xSero:
#319 has merged (`0824605d`) and rewrote `ConversationView.swift` substantially, so this will need a rebase. The Rust half of #306 is now in — `include_turns=false` is authoritative and page size is 20 — so what remains here is the scroll integration itself: prefetch, the "Latest" jump, and auto-collapse. Still wanted.

DatScreamer:
This PR is superseded by #331 (`feat(home): infinite scroll + pull-to-refresh for the home sessions list`). The conversation older-turns infinite scroll work is now carried on the `feat/home-sessions-infinite-scroll` branch (commit `d3ab5a22`), which this PR's home branch already contains. Keeping both open would race on the same `AppModel.swift` paging code and Rust cursor state. Recommend closing this one in favor of #331.

PR #338, opened 2026-09-13
bridge: reconcile agent probe metadata + render-stall follow-ups

0xSero:
## Summary

Post-#319 follow-ups, four logical commits:

1. **`bridge: reconcile agent probe metadata against the built-in catalog`** — `agent_catalog::reconcile_probe_metadata` folds a host's probe response into what Litter already knows before caching: `supports_ssh_bridge` is re-stated from the built-in catalog (the host manifest flag means "could an alleycat host bridge it", not "litter can launch it over SSH" — taking it verbatim enabled picker rows that fail on connect), and missing `presentation`/`capabilities` blocks fall back to the seeded entry instead of erasing known agents. Wired into `MobileClient`'s probe upsert. Includes unit tests.
2. **`build: compile the ios-dev profile at opt-level 2`** — the fast device lane keeps `debug=1` line tables but runs interpreter-side hot loops (streaming snapshot assembly, conversation projection, JSON parsing) an order of magnitude faster.
3. **`ios: gate expensive derivations behind snapshot revision`** — conversation projection, sessions derivation, and the PiP active-session computation now short-circuit on unchanged snapshot revisions instead of re-running deep equality/sort/group passes on every coalesced snapshot bump (~8 fps while streaming) and on duplicate binds.
4. **`android: stabilize composer callbacks and persist drafts immediately`** — `ConversationScreen` remembers its ComposerBar lambdas so argument-level skipping survives recomposition; composer draft persistence stays an immediate unconditional write (debounced variants lose typed text on quick navigation/backgrounding).

## Verification

- `make rust-test`: 787 passed / 0 failed (host)
- Mobile CI runs android + ios builds on this branch


0xSero:
The iOS job wasn't a code failure — it was cancelled at the 60-minute job timeout (`The job has exceeded the maximum execution time of 1h0m0s`). Everything inside the job passed: `** TEST EXECUTE SUCCEEDED **`, all test suites 0 failures, and the launch smoke test printed `App is alive`.

Root cause: the `SCCACHE_R2_*` secrets are not configured in the repo, so the guard step disables sccache and every iOS run cold-builds the Rust tree. Green PRs land at 34-41 minutes; this run was slower (~61 min) and got killed during post-cleanup just past the limit. Raised the ios job `timeout-minutes` to 90 so cold builds can't be cut off. (Adding the R2 sccache secrets back would cut iOS CI to roughly the 35-minute range — noted for follow-up.)

PR #346, opened 2026-09-13
bridge: restore fork lineage on cold thread lists

0xSero:
Revival of the closed #218 (branch deleted). Fixes the cold-`thread/list` half of #150: the fork rollout survived process restart and `thread/read` returned its parent, but cold `thread/list` discarded `forkedFromId` in the local thread-store list projection. The patch restores lineage from the rollout session metadata.

**Verification:**
- `git apply --check` of `patches/codex/thread-list-fork-lineage.patch` against a clean export of the pinned codex (`13595c36`) — applies cleanly; the submodule pin has not moved since the original draft.
- Sync-codex wiring (Makefile resolved to main's wildcard patch discovery; the patch is picked up automatically).
- Full physical-device acceptance from #150 (fork → follow up → leave → force restart → reopen from Sessions/Forks → continue on both iPhone This-Device and remote-VM paths) remains the issue's acceptance gate.


0xSero:
Follow-ups from the review, pushed as `411e6a26..1e4239c9`:

1. **CI now runs the new test.** The shared job applies the patch set (`sync-codex.sh --recorded-gitlink`) and then runs `cargo test -p codex-thread-store`, so `cold_list_threads_preserves_fork_lineage_from_rollout` executes on every PR instead of only locally.

2. **Per-thread re-reads removed.** Lineage is now carried through the rollout list projection instead of reopening each rollout file in `list_threads`: `ThreadItem` gains `forked_from_id` (filled from session meta during the head read), `stored_thread_from_rollout_item` forwards it, and `fill_missing_thread_item_metadata` merges state-DB metadata values in. `thread-list-fork-lineage.patch` was regenerated accordingly. The whole patch sequence was verified by resetting the submodule pristine and re-applying every patch from the committed files — the re-applied tree is byte-identical to the tested state.

3. **Self-parents / cycles / missing parents.** Left as-is intentionally: upstream fork always mints a fresh thread id, and with the projection path the parent id comes straight from the fork's own session meta. A missing parent (deleted, archived, or on another page) yields a fork entry whose parent id may not be present in the same page — same exposure as the previous per-read approach; deeper UX handling is tracked with #150.

4. **Silent error nit** — evaporated with the re-read block: `list_threads` no longer does per-thread file I/O for lineage.

5. **Test gaps.** The cold-list test now also asserts a non-fork thread keeps `forked_from_id: None`. The state-db-only and id-mismatch gaps are moot — no per-thread file reads remain, and the projection is per-file by construction. A mobile-side reducer test asserting a cold list result sets `is_fork` in the session summary is still open; noted on #150.

Verified locally: `cargo test -p codex-thread-store` 73/73 pass, `cargo check -p codex-rollout --all-targets` clean.

PR #351, opened 2026-09-13
feat(home): infinite scroll + pull-to-refresh for the home sessions list (carried from #331)

0xSero:
Carried from #331 so it can merge against current main: the original head branch lives in @DatScreamer's fork and could not be rebased here. All 15 home-sessions feature commits are cherry-picked with DatScreamer's authorship preserved (`-x` references the fork originals).

**What's included (from #331):** infinite scroll + pull-to-refresh on the home sessions list for both platforms, shared Rust cursor-paged session loading (`AppStore::load_threads_page`, retained per-runtime cursors, `sessionListHasMore` projection), first-page load on connect/reconnect, default page size 10→20, session-card spacing fixes, and the subagent-session exclusion (client-side filter).

**Dropped from the original branch, with reasons:**
- Its copy of the conversation older-turns pagination — #347 carries that feature (the revival of #308).
- Infra fix commits (detached-transport compilation, pinned-bridge builds, daemon pin alignment, mobile CI compat) — superseded by work already merged (#297) and the open #346/#350.
- The alleycat pin bump inside the subagent-exclusion commit — the production pin in #350 already contains the opencode subagent fix (`d503bfc` is an ancestor of the pinned revision); the client-side filter is kept and pins stay untouched here.

**Verification:** Rust test suite + Android unit tests via CI on this PR; the original feature was device-verified by @DatScreamer on both platforms (see #331).

Closes #331.


0xSero:
Reviewed for the next stability release. Closing this version without merging: the feature is useful, but the current implementation adds startup work and has correctness gaps that conflict with the release's latency requirements.

- `HomeDashboardModel.refreshAll` starts an unstructured Task and returns synchronously. The async refresh-control callback therefore ends the spinner before the network refresh finishes, and refreshes can overlap.
- Rust post-connect warmup, platform reconnect, and home activation can all request pages. The platform guards do not serialize the Rust warmup with cursor reads/writes, allowing an older first-page result to overwrite a later cursor.
- iOS home adds every loaded session as a retained hosting controller; growing this list is pagination, not view virtualization.
- Android removes the existing pinch gesture in this change, unrelated to pagination.

The current CI passes, but its coverage does not establish these runtime behaviors. A replacement should have one shared cursor owner, awaited refresh completion, bounded visible rows, and reconnect/overlap tests. Conversation pagination already merged in #347 remains in the release.

